Issue #82, opened 2025-05-05
Muti-User Exchange


AK2000:
### Describe the Request

In order to host an exchange for agents, there must be a mechanism to secure agents listening on the exchange. Multiple users should be able to use the exchange to communicate, but agents should only be able to communicate with other agents if they have the right permissions. A basic multi-user exchange will require:

- [ ] Authentication (presumably through Globus Auth)
- [ ] Mechanism for publishing/sharing agents
- [ ] Rate and message size limiting

Other potential features could include:

- Credential delegation for LLM agents
- Enhanced discoverability
- Different models for sharing agents (i.e. invocation, delegation, migration)
- Quota or budget model

gpauloski:
@AK2000, I've updating the issue to be the "parent" issue that tracks all issues related to the new exchange.

PR #86, opened 2025-05-08
Add client authentication to HTTP exchange

AK2000:
# Description
Add Globus authentication and TLS option to http exchange. Refactor code to place http exchange and all authentication related files into a subfolder. Note that this PR is not defining any concept of permissions, only authentication. At this point, permissions to the exchange can be enforced by setting policies on the Globus application (i.e. only allow uchicago.edu domains). Permissions are deferred because it raises many design discussions over what permissions we need to enable and how to enforce them.

### Fixes
- Begins to address #82 

### Type of Change
- [x] Refactoring (internal implementation changes)

## Testing
In addition to the tests provided an http exchange was stood up locally, registered as a Globus app, and run with a minimal script.

## Pull Request Checklist


AK2000:
pre-commit.ci autofix

academy/exchange/cloud/client.py
```diff
@@ -1,656 +1,377 @@
...
+            with the exchange. Used for passing the Globus bearer token
+        scheme: HTTP scheme, non-protected "http" by default.
+    """
+
+    def __init__(
+        self,
+        host: str,
+        port: int,
+        additional_headers: dict[str, str] | None = None,
+        scheme: str = 'http',
```
gpauloski: Can this be only `"http"` or `"https"`? If so, let's annotate it as `Literal['http', 'https']`.

academy/exchange/cloud/client.py
```diff
@@ -1,656 +1,377 @@
...
+
+class HttpExchange(ExchangeMixin):
+    """Http exchange client.
+
+    Args:
+        host: Host name of the exchange server.
+        port: Port of the exchange server.
+        additional_headers: Any other information necessary to communicate
+            with the exchange. Used for passing the Globus bearer token
+        scheme: HTTP scheme, non-protected "http" by default.
```
gpauloski, line 48: Can you add a note on `ssl_verify`? I assume that's just for validating the server certificate, in which case we should probably note that it should be true if using https.

academy/exchange/cloud/client.py
```diff
@@ -1,656 +1,377 @@
...
+) -> Generator[HttpExchange]:
+    """Context manager that spawns an HTTP exchange in a subprocess.
+
+    This function spawns a new process (rather than forking) and wait to
+    return until a connection with the exchange has been established.
+    When exiting the context manager, `SIGINT` will be sent to the exchange
+    process. If the process does not exit within 5 seconds, it will be
+    killed.
+
+    Note: The exclusion of authentication and ssl configuration is
```
gpauloski: Probably worth making this a warning block. Also, it might need to be indented to render as an admonition (not sure).
```
Warning:
    The exclusion...
```

academy/exchange/cloud/config.py
```diff
@@ -0,0 +1,154 @@
...
+    from typing_extensions import Self
+
+from pydantic import BaseModel
+from pydantic import ConfigDict
+from pydantic import Field
+
+if sys.version_info >= (3, 11):  # pragma: >=3.11 cover
+    import tomllib
+else:  # pragma: <3.11 cover
+    import tomli as tomllib  # type: ignore[import-not-found,unused-ignore]
```
gpauloski, line 28: Probably need to add the following to the `pytproject.toml` requirements (something else might already require it, but I'd rather be explicit).
```
"tomli ; python_version<'3.11'",
```
and also to the mypy additional requirements in `.pre-commit.config.yaml` so it can get type checked.

academy/exchange/cloud/config.py
```diff
@@ -0,0 +1,154 @@
...
+    Attributes:
+        host: Network interface the server binds to.
+        port: Network port the server binds to.
+        certfile: Certificate file (PEM format) use to enable TLS.
+        keyfile: Private key file. If not specified, the key will be
+            taken from the certfile.
+        auth: Authentication configuration.
+        log_file: Location to write logs.
+        log_level: Verbosity of logs.
+        max_message_bytes: Maximum size in bytes of messages received by
```
gpauloski: Can remove this and the associated attribute.

academy/exchange/cloud/login.py
```diff
@@ -0,0 +1,234 @@
...
+from globus_sdk.globus_app import GlobusApp
+from globus_sdk.globus_app import GlobusAppConfig
+from globus_sdk.globus_app import UserApp
+from globus_sdk.login_flows import CommandLineLoginFlowManager
+from globus_sdk.scopes import ScopeBuilder
+from globus_sdk.tokenstorage import SQLiteTokenStorage
+from globus_sdk.tokenstorage import TokenValidationError
+
+# Registered `ProxyStore Application` by jgpauloski@uchicago.edu
+ACADEMY_GLOBUS_CLIENT_ID = '1624cf3f-45ee-4f54-9de4-2d5d79191346'
```
gpauloski, line 24: Is this a new app that you've registered? If so, can you update the comments, the env names to be prefixed with `ACADEMY`, and the `ScopeBuilder` to use this variable as the client ID.

academy/exchange/cloud/login.py
```diff
@@ -0,0 +1,234 @@
...
+
+def is_client_login() -> bool:
+    """Check if Globus client identity environment variables are set.
+
+    Based on the Globus Compute SDK's
+    [`is_client_login()`](https://github.com/funcx-faas/funcX/blob/8f5b59075ae6f8e8b8b13fe1b91430271f4e0c3c/compute_sdk/globus_compute_sdk/sdk/login_manager/client_login.py#L24-L38){target=_blank}.
+
+    Returns:
+        `True` if `PROXYSTORE_GLOBUS_CLIENT_ID` and \
+        `PROXYSTORE_GLOBUS_CLIENT_SECRET` are set.
```
gpauloski: Update these env names.

academy/exchange/cloud/server.py
```diff
@@ -0,0 +1,400 @@
+"""HTTP message exchange client and server.
+
+To start the exchange:
+```bash
+python -m academy.exchange.http --host localhost --port 1234
```
gpauloski: These examples will need to be updated. Right now it would be 
```
python -m academy.exchange.cloud.server
```
But I think it'd be better to be `academy.exchange.cloud`. You can do this by moving
```python
if __name__ == '__main__':
    raise SystemExit(_main())
```
at the end of the file to a new file named `academy/exchange/cloud/__main__.py` file that has:
```python
from academy.exchange.cloud.server import _main

if __name__ == '__main__':
    raise SystemExit(_main())
```

academy/exchange/cloud/server.py
```diff
@@ -0,0 +1,400 @@
...
+    """
+
+    @middleware
+    async def authenticate(
+        request: Request,
+        handler: Callable[[Request], Awaitable[Response]],
+    ) -> Response:
+        loop = asyncio.get_running_loop()
+        try:
+            ## Needs to be run in executor because globus client is blocking
```
gpauloski: Is the globus client used internally in the authenticator threadsafe?

testing/constant.py
```diff
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-TEST_CONNECTION_TIMEOUT = 5
+TEST_CONNECTION_TIMEOUT = 15
```
gpauloski, line 3: Is there a test that takes longer than 5 seconds to run?

academy/exchange/cloud/login.py
```diff
@@ -0,0 +1,234 @@
...
+"""
+
+from __future__ import annotations
+
+import os
+import pathlib
+from typing import Any
+from typing import Literal
+
+import click
```
gpauloski, line 13: We'll need to add click and globus_sdk as requirements. I'm also wondering if maybe the cloud exchange requirements should be an optional dependency set in the pyproject?

gpauloski (changes requested):

academy/exchange/cloud/login.py
AK2000: Updated the names, but the client ID used by the library is different than the client ID of the exchange that is needed for the scope builder.

testing/constant.py
AK2000: My laptop is quite slow, but the `test_spawn_http_exchange` test in tests/exchange/cloud/client.py takes longer than 5 seconds to run for me

academy/exchange/cloud/login.py
AK2000: I added the requirements, but I don't think it's worth making a separate optional dependencies because these are required dependencies of proxystore, which is already a required dependency.

academy/exchange/cloud/server.py
AK2000: It is not. Since we only use a single method from the auth client, I think just protecting that method with a lock will make the authenticator thread safe. I also added notes in the doc strings of the protocol about this requirement.

gpauloski (approved):

gpauloski:
Looks good. I rebased and squashed the commits. One file had the wrong line endings, not sure how that got past pre-commit.

Pre-commit seems to be down, but can merge whenever it comes back.

PR #104, opened 2025-05-15
Add basic mailbox ownership model for `HttpExchange`

AK2000:
# Description
Currently all authenticated users can read and write from any mailbox. This PR implements a basic ownership model, so only agents created by the same client can discover and communicate with each other. This is implemented by passing around a bearer token when starting agents (not a good pattern). 

### Type of Change

## Testing
Tests added to verify

## Pull Request Checklist

### DIFF
--- a/academy/exchange/cloud/server.py
+++ b/academy/exchange/cloud/server.py
@@ -67,9 +67,9 @@ logger = logging.getLogger(__name__)
 BehaviorT = TypeVar('BehaviorT', bound=Behavior)
 _OKAY_CODE = 200
 _BAD_REQUEST_CODE = 400
+_UNAUTHORIZED_CODE = 401
 _FORBIDDEN_CODE = 403
 _NOT_FOUND_CODE = 404
-_UNAUTHORIZED_CODE = 405
 
 
 class _MailboxManager:
@@ -87,7 +87,7 @@ class _MailboxManager:
 
     def check_mailbox(self, client: str | None, uid: EntityId) -> bool:
         if not self.has_permissions(client, uid):
-            raise UnauthorizedError(
+            raise ForbiddenError(
                 'Client does not have correct permissions.',
             )
 
@@ -100,7 +100,7 @@ class _MailboxManager:
         behavior: tuple[str, ...] | None = None,
     ) -> None:
         if not self.has_permissions(client, uid):
-            raise UnauthorizedError(
+            raise ForbiddenError(
                 'Client does not have correct permissions.',
             )
 
@@ -113,7 +113,7 @@ class _MailboxManager:
 
     async def terminate(self, client: str | None, uid: EntityId) -> None:
         if not self.has_permissions(client, uid):
-            raise UnauthorizedError(
+            raise ForbiddenError(
                 'Client does not have correct permissions.',
             )
 
@@ -142,7 +142,7 @@ class _MailboxManager:
 
     async def get(self, client: str | None, uid: EntityId) -> Message:
         if not self.has_permissions(client, uid):
-            raise UnauthorizedError(
+            raise ForbiddenError(
                 'Client does not have correct permissions.',
             )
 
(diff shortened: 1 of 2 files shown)


academy/exchange/cloud/server.py
```diff
@@ -74,36 +74,64 @@
...
+
         return uid in self._mailboxes
 
     def create_mailbox(
         self,
+        client: str | None,
         uid: EntityId,
         behavior: tuple[str, ...] | None = None,
     ) -> None:
+        if not self.has_permissions(client, uid):
```
gpauloski, line 102: What's the purpose of this check before a mailbox has been created?

academy/exchange/cloud/server.py
```diff
@@ -151,7 +189,14 @@ async def _create_mailbox_route(request: Request) -> Response:
             text='Missing or invalid mailbox ID',
         )
 
-    manager.create_mailbox(mailbox_id, behavior)
+    client_id = request.headers.get('client_id', None)
+    try:
+        manager.create_mailbox(client_id, mailbox_id, behavior)
+    except UnauthorizedError:
+        return Response(
+            status=_UNAUTHORIZED_CODE,
```
gpauloski: I think I missed this in the last PR, but `_UNAUTHORIZED_CODE` should actually be 401 (right now its 405, which is the method not allowed error code). And in this context I think `_FORBIDDEN_ERROR` (403) is actually the correct code to return (as the user _is_ authorized but doesn't have the right permissions). https://en.wikipedia.org/wiki/List_of_HTTP_status_codes

academy/exchange/cloud/server.py
```diff
@@ -297,7 +374,9 @@ async def authenticate(
                 text='Missing required headers.',
             )
 
-        request['client_id'] = str(client_uuid)
+        headers = request.headers.copy()
+        headers['client_id'] = str(client_uuid)
+        request = request.clone(headers=headers)
```
gpauloski, line 379: Is it possible to avoid this clone? I'm wondering if it makes a copy of the JSON payload (which could potentially be large)?

tests/exchange/cloud/server_test.py
```diff
@@ -234,79 +253,193 @@ async def test_null_auth_cli() -> None:
         assert await response.text() == 'Unknown mailbox ID'
 
 
-@pytest.mark.asyncio
-@mock.patch('globus_sdk.ConfidentialAppAuthClient.oauth2_token_introspect')
-async def test_globus_auth_cli(
-    mock_token_response,
-) -> None:
+@pytest_asyncio.fixture
+async def auth_cli() -> AsyncGenerator[TestClient[Request, Application]]:
```
gpauloski: Does this func name mean auth CLI or auth client?

gpauloski (changes requested):
Looks good. There's a couple small questions but I think the only major change is the exact HTTP error codes used.

We might want to consider in a later PR adjusting the `Exchange` protocol to have more error types for forbidden/unauthorized errors and parsing the aiohttp errors into academy library errors in the `HttpExchange` client.

academy/exchange/cloud/server.py
AK2000: Creating a mailbox is idempotent, so there is no "mailbox exists" error but in the current implementation, you can't claim ownership or create a mailbox with the same id as a mailbox created by someone else, so this check is necessary. We could also create the id only for the owner of the mailbox, but I thought that choice would limit future sharing of agents.

academy/exchange/cloud/server.py
AK2000: This is necessary because request.headers is not assignable or mutable. I don't think this is a problem because only a shallow copy of the request is created. The body of the request will not be copied. And only a shallow copy of the request.headers will be created.

academy/exchange/cloud/server.py
gpauloski: Got it. I've wondered about introducing a `MailboxExistsError` but maybe a discussion for another day.

gpauloski (approved):

gpauloski:
I opened #105 so we can revisit updating the `Exchange` protocol to support more error types (like the unauthorized/forbidden errors the `HttpExchange` needs).

PR #115, opened 2025-06-04
Unbound exchange protocol

AK2000:
# Description
Changing exchange to be tightly coupled with mailbox and multiplexer

### Type of Change
- [x] Refactoring (internal implementation changes)

## Pull Request Checklist


academy/exchange/__init__.py
```diff
@@ -211,46 +335,62 @@ def get_handle(
...
+            handle.close(wait_futures=False)
+
+    def _message_handler(self, message: Message) -> None:
+        if isinstance(message, get_args(RequestMessage)):
+            self._handle_request(message)
+        elif isinstance(message, get_args(ResponseMessage)):
+            try:
+                handle = self.bound_handles[message.label]
+            except KeyError:
+                pass
```
gpauloski: Am I correct in understanding that this `KeyError` could happen if a handle is removed before receiving a response message? I'm wondering if we should do something here like log a warning?

academy/exchange/__init__.py
```diff
@@ -107,71 +235,67 @@ def discover(
...
-            BadEntityIdError: if a mailbox for `uid` does not exist.
+            MailboxClosedError: if the mailbox was closed.
+            TimeoutError: if a `timeout` was specified and exceeded.
         """
         ...
 
-    def send(self, uid: EntityId, message: Message) -> None:
-        """Send a message to a mailbox.
+    @abc.abstractmethod
+    def close(self) -> None:
```
gpauloski, line 289: Is there a requirement/suggestion on what order this should be called in subclasses? I noticed most implementations call `super().close()` first so wondering if we should note that in the docstring.

academy/exchange/cloud/client.py
```diff
@@ -1,382 +1,415 @@
...
+    def close(self) -> None:
+        """Close this exchange client."""
+        super().close()
+
+        self._session.close()
+        logger.debug('Closed exchange (%s)', self)
+
+    def status(self, mailbox_id: EntityId) -> MailboxStatus:
+        """Check status of mailbox on exchange."""
+        # TODO: Do we need to create a new endpoint to check status?
```
gpauloski: I think that would be good. Let's open an issue for this and update the TODO to mention the issue number so we don't lose track of it.

academy/exchange/hybrid.py
```diff
@@ -95,8 +91,89 @@ def __init__(
...
+        *,
+        start_listener: bool,
+    ) -> BoundHybridExchangeClient:
+        return BoundHybridExchangeClient(
+            self,
+            mailbox_id=mailbox_id,
+            name=name,
+            handler=handler,
+            start_listener=start_listener,
+            port=None if self._ports is None else next(self._ports),
```
gpauloski, line 112: Was the iterable of ports added for testing reasons?

academy/manager.py
```diff
@@ -128,23 +116,15 @@ def __str__(self) -> str:
...
 
     @property
-    def exchange(self) -> Exchange:
+    def exchange(self) -> BoundExchangeClient:
         """Exchange interface."""
         return self._exchange
 
     @property
-    def mailbox_id(self) -> ClientId:
+    def mailbox_id(self) -> EntityId:
```
gpauloski: I presume this was changed becuase `self._exchange.mailbox_id` returns `EntityID`? Let's keep it as `ClientId` and assert the type in `__init__` when we read it from the exchange.

academy/exchange/hybrid.py
AK2000: I used it for testing, but conceivably you may need it for some other policies (i.e. you want to specify that the direct messaging happens only on a specific port so you can open that port to inbound connections)

academy/exchange/cloud/client.py
AK2000: Created issue #118 and mentioned it in the TODO

academy/exchange/hybrid.py
AK2000: Was there a change that should be made to address this?

AK2000:
> The one big thing I wanted to raise for the discussion is the renaming of `UnboundExchangeClient` and `BoundExchangeClient`. I've been thinking that `ExchangeFactory` and `ExchangeClient`, respectively, might make more sense since we essentially have a factory patter and it would be less verbose. What do you think?
I agree, I thought the names were getting unwieldy and this is a better naming scheme. I renamed the symbols

academy/exchange/hybrid.py
gpauloski: Maybe just alter the docstring for the parameter to make it more clear that the purpose of the iterable is to give each client a unique port from a user-defined set and that a `StopIteration` exception would be raised in `bind_*` methods if the number of clients in the process is greater than the length of the iterable.

gpauloski (approved):
Left one last comment on the docstring of the `ports` parameter, but otherwise looks great!

Issue #107, opened 2025-05-19
Passing of Delegated Tokens When Launching Agents


AK2000:
### Describe the Request

The current method of instantiating an agent that can use the authenticated exchange is not ideal. It relies on passing the same access token to all launch agents so they can communicate with the exchange. In this model these tokens are subject to expiration, with no recourse for the launched agent, they may be intercepted and used without protection, and they can only be revoked for all launched agents. The canonical flow would involve creating a unique auth client for each agent, and creating delegated tokens to each launched agent.

PR #124, opened 2025-06-17
Add agent registration info protocol

AK2000:
# Description
Introduce a protocol so that exchanges can return additional information when registering an Agent and get passed additional information when launching an agent. This protocol is the beginning of implementing #107 

### Type of Change

- [ ] New feature (non-breaking change which adds functionality)
- [ ] Refactoring (internal implementation changes)
- [ ] Documentation update (no changes to the code)
- [ ] CI change (changes to CI workflows, packages, templates, etc.)
- [ ] Version changes (changes to the package or dependency versions)

## Pull Request Checklist

### DIFF
--- a/academy/agent.py
+++ b/academy/agent.py
@@ -97,7 +97,7 @@ class Agent(Generic[BehaviorT]):
 
     Args:
         behavior: Behavior that the agent will exhibit.
-        agent_id: EntityId of this agent in a multi-agent system.
+        agent_info: RegistrationInfo of this agent created by the exchange.
         exchange: Message exchange of multi-agent system. The agent will close
             the exchange when it finished running.
         config: Agent execution parameters.
--- a/academy/agent.py
+++ b/academy/agent.py
@@ -70,8 +70,8 @@ class AgentRunConfig:
 def _agent_trampoline(
     behavior: BehaviorT,
     agent_id: AgentId[BehaviorT],
-    exchange: ExchangeFactory[AgentRegistrationT],
     agent_info: AgentRegistrationT,
+    exchange: ExchangeFactory[AgentRegistrationT],
     config: AgentRunConfig,
 ) -> Agent[BehaviorT]:
     return Agent(
@@ -100,10 +100,10 @@ class Agent(Generic[BehaviorT]):
 
     Args:
         behavior: Behavior that the agent will exhibit.
-        agent_id: EntityId of this agent in a multi-agent system.
+        agent_id: Unique ID of this agent in a multi-agent system.
+        agent_info: Registration info for this agent returned by the exchange.
         exchange: Message exchange of multi-agent system. The agent will close
             the exchange when it finished running.
-        agent_info: RegistrationInfo of this agent created by the exchange.
         config: Agent execution parameters.
     """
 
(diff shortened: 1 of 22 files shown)


academy/agent.py
```diff
@@ -107,14 +107,14 @@ def __init__(
         self,
         behavior: BehaviorT,
         *,
-        agent_id: AgentId[BehaviorT],
+        agent_info: RegistrationInfo[BehaviorT],
         exchange: ExchangeFactory,
         config: AgentRunConfig | None = None,
     ) -> None:
-        self.agent_id = agent_id
+        self.agent_info = agent_info
```
gpauloski, line 120: It might be worth keeping an `agent_id` attribute as a property since it is accessed so many times (would also reduce the loc changed).

```python
@property
def agent_id(self) -> AgentId[BehaviorT]:
    return self.agent_info.agent_id
```

academy/exchange/__init__.py
```diff
@@ -54,6 +57,27 @@ class MailboxStatus(enum.Enum):
     """Mailbox was terminated and no longer accepts messages."""
 
 
+class RegistrationInfo(Protocol[BehaviorT]):
```
gpauloski: Will we ever need to have a similar class for user registrations? If not `AgentRegistration` may be a better name.

academy/exchange/__init__.py
```diff
@@ -74,11 +98,12 @@ def _create_transport(
         mailbox_id: EntityId | None = None,
         *,
         name: str | None = None,
+        registration_info: RegistrationInfo[BehaviorT] | None = None,
     ) -> ExchangeTransport: ...
 
     def create_agent_client(
         self,
-        agent_id: AgentId[BehaviorT],
+        agent_info: RegistrationInfo[BehaviorT],
```
gpauloski: I assume that a `FooExchangeFactory.create_agent_client()` will need to take the exact `RegistrationInfo` type produced by `FooExchangeTransport.register_agent()`. We'll probably need to make `ExchangeTransport` and `ExchangeFactory` generic on the `RegistrationInfo` type then otherwise mypy is going to complain as soon as we add something that's not `BasicRegistrationInfo`.

academy/exchange/__init__.py
```diff
@@ -398,13 +427,13 @@ def register_agent(
         Returns:
             Agent ID.
```
gpauloski: Update this to be Agent registration information.

academy/exchange/cloud/server.py
```diff
@@ -10,8 +10,8 @@
 from academy.exchange.cloud.client import HttpExchange
 
 with HttpExchange('localhost', 1234) as exchange:
-    aid = exchange.register_agent()
-    mailbox = exchange.get_mailbox(aid)
+    agent_info = exchange.register_agent()
+    mailbox = exchange.get_mailbox(agent_info.agent_id)
```
gpauloski: Oops this code block is very out of date now.

academy/launcher.py
```diff
@@ -43,7 +44,7 @@ def _run_agent_on_worker(agent: Agent[Any]) -> None:
 @dataclasses.dataclass
 class _ACB(Generic[BehaviorT]):
     # Agent Control Block
-    agent_id: AgentId[BehaviorT]
+    agent_info: RegistrationInfo[BehaviorT]
```
gpauloski: Similar to with `Agent`, for ease of use let's keep an `agent_id` attribute and a separate `info`/`registration` attribute.

gpauloski (changes requested):

academy/exchange/__init__.py
AK2000: I implemented this fix by separating the `AgentId` and the `AgentRegistration`. Parameterizing the `ExchangeFactory` and `ExchangeTransport` by the registration type also has the nice property that it tightly couples the two.

PR #128, opened 2025-06-25
Rewrite Academy in asyncio

### LINKED ISSUE #100: Switch to async model for agent communication
Concurrent communication is currently handled entirely through threads. This makes debugging a pain and is generally better suited for asyncio. However, we don't want to force user `Behavior`s into be async (but it would be nice to support both).

Switching to an async model would require the following changes:
- `Exchange` methods becomes async. Not too tricky because the exchange methods are really only used internally within other interfaces---just lots of code to touch.
- `Agent` becomes async. Again not to tricky because its all internal code and makes a lot of the threading headaches go away. Actions can still be submitted to a thread pool with `asynio.run_in_executor()` and we could more easily support sync and async actions/loops.
- `Handle` needs to internally call the async methods of the `Exchange` while still being a synchronous interface. This is the trickier one. We may need to follow Dask's stra for calling async code [distributed/distributed/utils.py at main · dask/distributed](https://github.com/dask/distributed/blob/main/distributed/utils.py#L444).

gpauloski:
# Description

This is a rewrite of Academy to use asyncio, which is generally more suitable for the I/O bound, asynchronous programming model that Academy uses. (This is a followup to #119 which did a fair amount of prep work for this PR.)

The underlying motivation for this change is to avoid the significant use of background threads for communication which were a headache for debugging and setup/teardown of resources.

Summary of changes:
* Agents, handles, exchanges, launcher, manager are now async
* Low-level socket utilities are consolidated to `academy.socket` and all async now
* Listener/server threads in exchange transports and clients are now asyncio Tasks that are cancelled when resources are closed.
* Tests are now much faster without starting/joining threads everywhere :)
* Agent:
  * Changed `terminate_on_exit` to `terminate_on_success`
  * Actions and control loops are executed as tasks in agent's event loop
* Handles:
  * Removed `mailbox_id` param of `BoundRemoteHandle` which can be inferred from `transport.mailbox_id`
  * Renamed `Handle.mailbox_id` to `Handle.client_id`
  * Renamed `BoundRemoteHandle` to `RemoteHandle`
* Launcher
  * Wrapper functions submitted to launchers now create an event loop to run the agent
  * asyncio Tasks are used to submit functions to executors and handle the results, greatly reducing the state the launcher needs to track in the ACBs
* Manager
  * Added `from_exchange_factory()` class method which registers a unique client for the manager

### Fixes

- Fixes #100

### Type of Change

- [ ] New feature (non-breaking change which adds functionality)
- [ ] Refactoring (internal implementation changes)
- [ ] Documentation update (no changes to the code)
- [ ] CI change (changes to CI workflows, packages, templates, etc.)
- [ ] Version changes (changes to the package or dependency versions)

## Testing

All tests updated and examples.

## Pull Request Checklist


x42005e1f:
Hello. I saw that you used the queues from my [aiologic](https://github.com/x42005e1f/aiologic) package in this PR. Thank you for your trust, but that led me to ask: why not [culsans](https://github.com/x42005e1f/culsans) queues? Unlike aiologic queues, they have full compatibility with standard Python 3.13+ queues (including the `shutdown()` method) and are covered by tests from the CPython repository.

Is the choice related to performance targeting? Or with the age of the last commit? Or is it about something else?

gpauloski:
Thanks for the detailed writeup! The discussion thread is very helpful and answered all of my questions.

I will give culsans a try :)

PR #146, opened 2025-07-02
Globus Exchange Client

### LINKED ISSUE #107: Passing of Delegated Tokens When Launching Agents

AK2000:
# Description
The globus exchange client implements the launching of agents as globus resource servers and the passing of delegated tokens for authentication of agents. To do this, the exchange client is tightly coupled with globus services and implements a `globus_sdk.BaseClient` to handle common complexities like storing and refreshing tokens and retrying and backing off requests.

### Fixes
- Fixes #107 

### Type of Change

## Pull Request Checklist

### DIFF
--- a/academy/exchange/cloud/authenticate.py
+++ b/academy/exchange/cloud/authenticate.py
@@ -74,9 +74,6 @@ class GlobusAuthenticator:
             Ignored if `auth_client` is provided.
         client_secret: Globus application client secret. See `client_id` for
             details. Ignored if `auth_client` is provided.
-        audience: Intended audience of the token. This should typically be
-            the resource server of the the token was issued for. E.g.,
-            the UUID of the Academy Exchange application.
         token_cache_limit: Maximum number of (token, identity) mappings
             to store in memory.
         token_ttl_s: Time in seconds before invalidating cached tokens.


academy/exchange/cloud/globus.py
```diff
@@ -0,0 +1,501 @@
...
+                'Launching child agents is\
+                                       currently not implemented.',
+            )
+
+        loop = asyncio.get_running_loop()
+        if self._auth_client is None:
+            # Need assert because globus_sdk annotations are not tight
+            assert isinstance(globus_sdk.AuthClient.resource_server, str)
+
+            # TODO: This may issues a request, but putting it inside an
```
AK2000: @gpauloski This is definitely worth taking a look at if you get a chance. `get_authorizer` may issue a request to Globus Auth, so it really should be run within an executor. However, putting it within the executor causes a race condition within the JSONTokenStorage leading to JSON Decoding errors. Leaving in like this seems to resolve those issues but may block the event loop. Since I couldn't resolve this after some effort, my thought was to leave this and open an issue.

AK2000:
pre-commit.ci autofix

academy/exchange/cloud/authenticate.py
```diff
@@ -117,7 +120,10 @@ def _token_introspect(
         self,
         token: str,
     ) -> globus_sdk.response.GlobusHTTPResponse:
-        return self.auth_client.oauth2_token_introspect(token)
+        response = self.auth_client.oauth2_token_introspect(token)
+        logger.info(f'Authenticated token with globus: {token}.')
+        logger.debug(f'Token introspect response: {response}.')
```
yadudoc, line 124: Is this log level intentional? It looks like the cloud config sets log_level to debug by default.

academy/exchange/cloud/globus.py
```diff
@@ -0,0 +1,606 @@
...
+from academy.exchange.transport import MailboxStatus
+from academy.identifier import AgentId
+from academy.identifier import EntityId
+from academy.identifier import UserId
+from academy.message import Message
+from academy.serialize import NoPickleMixin
+
+logger = logging.getLogger(__name__)
+
+ACADEMY_PROJECT = '6c357d85-dfb2-4f22-8ab3-dc617849aac5'
```
yadudoc: This doesn't seem to be used in this module. Is this imported elsewhere? If so it should probably go into `constants.py` or similar.

tests/unit/exchange/cloud/login_test.py
```diff
@@ -39,21 +39,16 @@ def mock_env_credentials() -> Generator[tuple[str, str], None, None]:
...
 def test_get_token_storage_academy_default(tmp_path: pathlib.Path):
     env = {
         'ACADEMY_HOME': str(tmp_path),
     }
     with mock.patch.dict(os.environ, env):
-        filepath = tmp_path / 'storage.db'
-        storage = get_token_storage()
-        assert filepath.is_file()
-        storage.close()
+        get_token_storage()
```
yadudoc: This test should assert something.

tests/integration/globus_launch_agents_test.py
```diff
@@ -0,0 +1,54 @@
...
+        super().__init__()
+
+    @action
+    async def echo(self, text: str) -> str:  # pragma: no cover
+        print('Running action')
+        return text
+
+
+@pytest.mark.skip(reason='Responses not mocked sufficiently.')
+async def test_full_globus_exchange_client() -> None:  # pragma: no cover
```
yadudoc: This should go into integration tests.

yadudoc (commented):
@AK2000 I've only had a chance to skim a part of this PR. I'd like to take a closer look at the `globus` module, but I don't want to hold up this PR. 

I've got mostly minor comments, the overall change here looks reasonable.

academy/exchange/cloud/authenticate.py
AK2000: Oh, I think the default should be info, but otherwise these two log levels were intentional

academy/exchange/cloud/globus.py
AK2000: Ope, this was a left over when I thought we could launch all the agents under a single project, but we don't do this since all the people launching agents would need manage_project permissions to that project.

tests/integration/globus_launch_agents_test.py
AK2000: It is? Or do you mean that it shouldn't be skipped?

PR #171, opened 2025-07-14
Change default behavior of handles to return final action result instead of awaitable future

### LINKED ISSUE #167: Change handle action to return final result instead of awaitable future
### Describe the Request

We’ve been discussing a possible (breaking) API change to improve ergonomics around how results are retrieved from async handle actions.

**Current Behavior**
Currently, invoking an action like `handle.get_count()` returns a future-like object, which must then be awaited to retrieve the result:

```python
future = await handle.get_count()
result = await future
# or
result = await (await handle.get_count())
```
This double-await pattern was confusing during a tutorial for those less familiar with async and leads to some aesthetically unpleasant when most users want the result immediately. This pattern is somewhat leftover from the prior sync interface which returned `concurrent.futures.Future`; however, the async model doesn't require use of futures here to achieve concurrency.

**Proposed Change**
We propose changing the default behavior so that:

```python
result = await handle.get_count()
```
directly returns the final result of the action. This would make immediate blocking behavior more concise and intuitive.

If users still want to manually manage the coroutine/future (e.g., schedule it concurrently), they can do so using standard asyncio tools:
```python
task = asyncio.create_task(handle.get_count())
# ... do other work ...
await task
print(task.result())
```

**Tradeoffs**
* This is a breaking change
* Shifts the default behavior to "await now" rather than "await later"
* Arguably improves ergonomics and readability for the common case

Based on the conversation in slack with @AK2000 and @yadudoc, we are in a general consensus that this is better default behavior and worth the breaking change.

### Sample Code

```python

```

gpauloski:
## Summary

Changes the default behavior of handles to return the final action result instead of an awaitable future. This improves ergonomics in the common case, without removing the possibility to later retrieve results (by creating a task).

**Previous Behavior**
```python
future = await handle.get_count()
result = await future
# or
result = await (await handle.get_count())
```

**New Behavior**
```python
# Blocking (common-case)
result = await handle.get_count()

# Non-blocking
task = asyncio.create_task(handle.get_count())
# ... do other work ...
await task
print(task.result())
```

## Related Issue

- Closes #167
## Changes

## Testing

Updated all of the tests. For the most part, updating the mypy plugin found all of the changes that needed to be made, except for those that call `Handle.action(name)` directly.

## Pull Request Checklist

### DIFF
--- a/docs/guides/migrating.md
+++ b/docs/guides/migrating.md
@@ -8,8 +8,8 @@ Academy v0.3 makes a few breaking changes to improve the ergonomics of common us
 
 ### Handle actions are blocking by default
 
-Previously, invoking an action on a [`Handle`][academy.handle.Handle.action] returns a [`Future`][asyncio.Future] to the result.
-This resulted in verbose syntax when the result was immediately needed.
+Previously, invoking an action on a [`Handle`][academy.handle.Handle.action] returned a [`Future`][asyncio.Future] to the result.
+This resulted in verbose syntax when the result was immediately needed:
 ```python
 future = await handle.get_count()
 result = await future
@@ -17,7 +17,7 @@ result = await future
 result = await (await handle.get_count())
 ```
 
-Now, action requests block and return the final result of the action.
+Now, action requests block and return the final result of the action:
 ```python
 result = await handle.get_count()
 ```
@@ -28,6 +28,9 @@ task = asyncio.create_task(handle.get_count())
 await task
 print(task.result())
 ```
+Using tasks is especially useful when launching multiple long-running actions concurrently and waiting for them in a flexible manner.
+For example, instead of waiting for each action sequentially, you can start them all at once and then wait for them to complete using [`asyncio.wait()`][asyncio.wait] or [`asyncio.as_completed()`][asyncio.as_completed].
+
 
 ## Migrating to v0.2
 


AK2000 (approved):
I think this is good. My only concern is when we release another breaking change. We might be a point where we should release it fast, so people don't develop with the old version and have to fix their code later

README.md
```diff
@@ -68,11 +68,10 @@ async with await Manager.from_exchange_factory(
 ) as manager:
     agent_handle = await manager.launch(SensorMonitorAgent)
 
-    future = await agent_handle.set_process_threshold(2.0)
-    await future
+    await agent_handle.set_process_threshold(2.0)
     await asyncio.sleep(5)
-    future = await agent_handle.get_last_reading()
-    value = await future
+    value = await agent_handle.get_last_reading()
```
yadudoc, line 73: Much cleaner!

docs/guides/migrating.md
```diff
@@ -1,15 +1,46 @@
...
+result = await future
+# or
+result = await (await handle.get_count())
+```
+
+Now, action requests block and return the final result of the action.
+```python
+result = await handle.get_count()
+```
+Code that wants to submit the request and later block on it can create a [`Task`][asyncio.create_task].
```
yadudoc, line 24: Would be good to add a note here on why someone might want to do this. For eg, launch several long-running actions and wait on them asynchronously with `asyncio.as_completed`.

yadudoc (approved):
The interface looks much nicer now! I expect we'll move fast and have breaking changes for a while. We should consider using scriv + changelog to make some of this easier.

PR #175, opened 2025-07-15
Add helper method for safely running sync functions within async actions

### LINKED ISSUE #93: Support deferred behavior initialization
Currently, users must initialize a behavior and submit that instance to be launched. This means the entire instance is serialized and shipped to the process that will execute the agent. To handle non-serializable data and deferred initialization, the `setup()` method is called from the execution process.

It would be good to also support submitting the behavior type and parameters to be launched, with the parameters be serialized and shipped instead. I think we can support both in a backwards compatible way.

```python
def launch(
    self,
    behavior: Behavior | Callable[P, Behavior], 
    *,
    args: P.args,
    kwargs: P.kwargs,
) -> ...:
```

### LINKED ISSUE #99: Deadlock when max action concurrency is reached
It's possible to get a deadlock if the action thread pool of an agent gets filled up with actions that are blocking, causing no other actions to get scheduled. We've already started to document that agents have a max concurrency, but maybe we need a mechanism to warn the user when that happens?

I'm not 100% sure what the best approach is. We could default to no max concurrency, support and encourage async actions, or return errors to actions when the current thread pool is overloaded.

gpauloski:
## Summary

Adds the `Agent.agent_run_sync()` helper method for safely running sync functions within async actions/loops/methods. This is essentially a wrapper around `loop.run_in_executor` with some extra features for supporting keyword arguments, configuring the max concurrency, and logging warnings when the thread pool executor is full.

Plenty of applications will want to call synchronous functions in their agents, such as third-party libraries, so I think it will be good to have a suggested mechanism for doing so.

If users want to customize the sync method calling, they can still use `loop.run_in_executor` manually. I considered adding an `executor` parameter to `Agent.agent_run_sync()` so agents could create their own executor (i.e., a process pool executor), but this breaks the paramspec typing for mypy and means we can't reliably raise warnings when the executor is overloaded.

We had previously discussed support both sync and async actions/loops. My idea for this was to have the runtime to schedule sync actions/loops in a thread pool using a mechanism similar to this, but I think the complexity of supporting multiple code paths is not worth it (i.e., sync/async variants of decorators and startup/shutdown in the runtime).

**Note:** I also removed some dead code related to threading from the v0.1/sync days.

## Related Issue

Closes #99

As actions are now async, there's no upper bound on the concurrent actions to be executed so the type of deadlock encountered in this issue is not possible. I had left the issue open in case we revisited sync actions, but as discussed above I think this approach of running sync code in a thread from an async action is better. I did, however, use the suggestion of logging warnings when the thread pool is overloaded.

## Changes

## Testing

Added new unit tests.

## Pull Request Checklist

### DIFF
--- a/academy/context.py
+++ b/academy/context.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import asyncio
 import dataclasses
+from concurrent.futures import ThreadPoolExecutor
 from typing import Any
 from typing import Generic
 from typing import TYPE_CHECKING
@@ -73,5 +74,7 @@ class AgentContext(Generic[AgentT]):
     """ID of the exchange as registered with the exchange."""
     exchange_client: AgentExchangeClient[AgentT, Any]
     """Client used by agent to communicate with the exchange."""
+    executor: ThreadPoolExecutor
+    """Thread-pool executor used for running synchronous tasks."""
     shutdown_event: asyncio.Event
     """Shutdown event used to signal the agent to shutdown."""
--- a/academy/agent.py
+++ b/academy/agent.py
@@ -8,7 +8,6 @@ import sys
 import warnings
 from collections.abc import Coroutine
 from collections.abc import Generator
-from concurrent.futures import Executor
 from datetime import timedelta
 from typing import Any
 from typing import Callable
@@ -94,6 +93,7 @@ class Agent:
 
     def __init__(self) -> None:
         self.__agent_context: AgentContext[Self] | None = None
+        self.__agent_run_sync_semaphore: asyncio.Semaphore | None = None
 
     def __repr__(self) -> str:
         return f'{type(self).__name__}()'
@@ -276,12 +276,10 @@ class Agent:
         self,
         function: Callable[P, R],
         /,
-        *args: Any,
-        executor: Executor | None = None,
-        timeout: float | None = None,
-        **kwargs: Any,
+        *args: P.args,
+        **kwargs: P.kwargs,
     ) -> R:
-        """Run a blocking function in an executor.
+        """Run a blocking function in separate thread.
 
         Example:
             ```python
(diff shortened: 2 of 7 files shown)


academy/context.py
```diff
@@ -73,5 +74,7 @@ class AgentContext(Generic[AgentT]):
     """ID of the exchange as registered with the exchange."""
     exchange_client: AgentExchangeClient[AgentT, Any]
     """Client used by agent to communicate with the exchange."""
+    executor: ThreadPoolExecutor
```
AK2000, line 77: Could we make this an arbitrary concurrent.futures executor? I'm thinking this would make the helper method more flexible to delegate blocking work not just to threads but to processes. It would also allow some "canonical" example of writing scalable agents.  I think there may be more things to be thought out here, so I might move this to an issue.

AK2000 (approved):
Looks good.

academy/context.py
gpauloski: I did initially write it in such a way that a user could, in the future, pick what the default executor type is for their agent, but I ended up reverting this. I think there's too much nuance around switching from a thread to process pool that it should not be a configuration someone can just flip; rather, it should be specific to the implementation of their agent (i.e., by creating their own `ProcessPoolExecutor` and calling `loop.run_in_executor()` instead of `agent_run_sync()`).

One immediate problem would be submitting sync methods of the agent would cause cause the entire `Agent` object to be serialized (because of the bound `self` parameter). Making `agent_run_sync()` always use a thread pool at least ensures it is always safe to submit a sync method that needs to access the agent's state.

I also took some inspiration from [`loop.set_default_executor()`](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.set_default_executor) which only allows thread pool executors because asyncio & third-party libraries assume the vent loop is always configured to use a thread pool executor (see: https://bugs.python.org/issue34075).

If we do end up supporting dispatching work to process pools, I think it should be through a separate mechanism that makes it more clear the potential hazards of multiprocessing.

PR #177, opened 2025-07-15
Read exchange client from context in handle

### LINKED ISSUE #169: Shared handle references in `Agent.__init__` do not get rebound
### Describe the problem.

Creating a "shared reference" to a handle in `Agent.__init__` means that the references to the original handle do not get replaced with the rebound handle.

Here's an example of a closure that captures a handle. If the closure is created in `__init__`, the unbound handle (or in this case, handle bound to the client because it's a threadpool executor) gets captured. But if the closure is created in an action after the handle binding has happened, then everything is okay.

```python

import asyncio
from concurrent.futures import ThreadPoolExecutor

from academy.agent import action
from academy.agent import Agent
from academy.exchange.local import LocalExchangeFactory
from academy.handle import Handle
from academy.manager import Manager


class Child(Agent):
    @action
    async def child(self) -> None:
        print('Running action in child')


def factory(handle: Handle[ExampleTool]):
    async def foo():
        return await handle.child()

    return foo


class Parent(Agent):
    def __init__(self, child: Handle[Child]):
        super().__init__()
        self.child = child
        self.tool = factory(self.child)

    @action
    async def parent_v1(self):
        tool = factory(self.child)
        result = await tool()
        print('Finished call to child using new tool')
        return result

    @action
    async def parent_v2(self):
        result = await self.tool()
        print('Finished call to child using instance attribute')
        return result


async def main() -> None:
    ) as manager:
        child = await manager.launch(Child)
        handle = await manager.launch(Parent, args=(child,))
        await handle.parent_v1()  # Works
        await handle.parent_v2()  # Hangs


if __name__ == '__main__':
    asyncio.run(main())
```

If a user is using a non-thread based executor, they would end up just getting an `UnboundHandle` exception rather than a silent hang which is better but not great as its still not clear why that would occur.

I wonder if the solution might be to get rid of `UnboundRemoteHandle` and make `RemoteHandle` mutable. In this case, `RemoteHandle` would have bound and unbound states and by mutating an instance to be bound we'd ensure all references to the same handle are valid.

@AK2000, what do you think?

```console
pip install academy-py
```

### Version

v0.2.0

### Python Version

All

### OS and Platform

All

AK2000:
## Summary
In a handle read the exchange client from a contextvar by default, using only the set exchange when the contextvar is not present. 

This change limits the need for re-binding handles based on the current use. Previously, we had to search for handles as properties of agents and bind them to the agent exchange. This got tricky very fast. We needed a custom type to search through lists or dictionaries of handles and couldn't find handles that were for instance passed into tools. A typical use case for a client handle will be inside of a context (either the exchange client context or the manager context). In both of these cases we can set the exchange client as a contextvar upon entering the client. A typical use case for an agent will only use a handle after an agent is started, where we set the contextvar.

The atypical use cases get messy. Particularly because an ExchangeClient must also keep track of the handles that are using if to send messages. If for instance a single handle is used to execute two actions with two different ExchangeClients, it will now belong to both of them and may be closed by both of them.

## Related Issue
- Fixes #169 

## Changes

## Pull Request Checklist

### DIFF
--- a/academy/handle.py
+++ b/academy/handle.py
@@ -404,21 +404,30 @@ class RemoteHandle(Generic[AgentT]):
         return (RemoteHandle, (self.agent_id, None))
 
     def __repr__(self) -> str:
-        return f'{type(self).__name__}(agent_id={self.agent_id!r})'
+        try:
+            exchange = self.exchange
+            client_id = self.client_id
+        except HandleNotBoundError:
+            exchange = None
+            client_id = None
+        return (
+            f'{type(self).__name__}(agent_id={self.agent_id!r}, '
+            f'client_id={client_id!r}, exchange={exchange!r})'
+        )
 
     def __str__(self) -> str:
         name = type(self).__name__
-        return f'{name}<agent: {self.agent_id}>'
+        try:
+            client_id = self.client_id
+        except HandleNotBoundError:
+            client_id = None
+        return f'{name}<agent: {self.agent_id}; mailbox: {client_id}>'
 
     def __getattr__(self, name: str) -> Any:
         async def remote_method_call(*args: Any, **kwargs: Any) -> R:
             return await self.action(name, *args, **kwargs)
 
         return remote_method_call
-
-    def clone(self) -> RemoteHandle[AgentT]:
-        """Create an unbound copy of this handle."""
-        return RemoteHandle(self.agent_id, None)
 
     async def _process_response(self, response: Message[Response]) -> None:
         future = self._futures.pop(response.tag, None)
--- a/academy/handle.py
+++ b/academy/handle.py
@@ -8,6 +8,7 @@ import time
 import uuid
 from collections.abc import Iterable
 from collections.abc import Mapping
+from contextvars import ContextVar
 from types import TracebackType
 from typing import Any
 from typing import Generic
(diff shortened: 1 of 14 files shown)


academy/handle.py
```diff
@@ -321,72 +322,10 @@ async def shutdown(self, *, terminate: bool | None = None) -> None:
...
-
-    async def ping(self, *, timeout: float | None = None) -> float:
-        """Raises [`HandleNotBoundError`][academy.exception.HandleNotBoundError]."""  # noqa: E501
-        raise HandleNotBoundError(self.agent_id)
-
-    async def shutdown(self, *, terminate: bool | None = None) -> None:
-        """Raises [`HandleNotBoundError`][academy.exception.HandleNotBoundError]."""  # noqa: E501
-        raise HandleNotBoundError(self.agent_id)
+exchange_context: ContextVar[ExchangeClient[Any] | None] = ContextVar(
+    'exchange_client',
```
gpauloski: Should this be `'exchange_context'` to match the var name?

academy/handle.py
```diff
@@ -418,6 +356,34 @@ def __init__(
...
+        Returns:
+            The ExchangeClient
+
+        Raises:
+            HandleNotBoundError: If the exchange client can't be found.
+
+        """
+        exchange = exchange_context.get()
+        if exchange is None and self._exchange is None:
+            raise HandleNotBoundError(self.agent_id)
```
gpauloski: Should we rename this error now that we've pretty much eliminated the notion of "binding" handles?

Maybe something along the lines of `ExchangeClientNotFoundError`, `ExchangeClientUnsetError`, `ExchangeClientMissingError`

academy/handle.py
```diff
@@ -399,14 +338,13 @@ class RemoteHandle(Generic[AgentT]):
 
     def __init__(
         self,
-        exchange: ExchangeClient[Any],
         agent_id: AgentId[AgentT],
+        exchange: ExchangeClient[Any] | None,
```
gpauloski: Can we make this default to `None`? I like the idea of being able to just instantiate a new handle inside an agent as long as you know the ID of the other agent:
```python
class Example(Agent):
    @action
    async def refer(self, agent_id: AgentId[Any]) -> None:
         self.handle = RemoteHandle(agent_id)
```

academy/manager.py
```diff
@@ -154,6 +155,9 @@ def __init__(
         logger.info('Initialized manager (%s)', self.user_id)
 
     async def __aenter__(self) -> Self:
+        self.exchange_context_token = exchange_context.set(
+            self.exchange_client,
+        )
```
gpauloski, line 160: We should recommend strongly that the `Manager` be used as a context manager in the docstring. For the most part, I think creating a `Manager` without a context manager will still work as handles created by the `Manager` get passed the correct exchange client, but if someone manually made a handle then they'd run into issues.

academy/handle.py
```diff
@@ -399,14 +338,13 @@ class RemoteHandle(Generic[AgentT]):
...
     ) -> None:
-        self.exchange = exchange
         self.agent_id = agent_id
-        self.client_id = exchange.client_id
+        self._exchange = exchange
```
gpauloski, line 346: If an exchange is provided, should we call `exchange.register_handle()`?

The relationship between `RemoteHandle.__init__`, `RemoteHandle.exchange`, `ExchangeClient.register_handle`, and `ExchangeClient.get_handle` is a bit confusing w.r.t. when a handle gets attached to its client. I'm sort of in favor of removing `ExchangeClient.get_handle` and having the `RemoteHandle` be entirely responsible for attaching itself to whatever client it ends up using.

academy/handle.py
```diff
@@ -418,6 +356,34 @@ def __init__(
         self._futures: dict[uuid.UUID, asyncio.Future[Any]] = {}
         self._closed = False
 
+    @property
+    def exchange(self) -> ExchangeClient[Any]:
```
gpauloski, line 368: Could this method be more simply written as the following?
```python
if self._exchange is not None:
    return self._exchange

exchange = exchange_context.get()
if exchange is None:
    raise HandleNotBoundError(self.agent_id)

self._exchange = exchange
self._exchange.register_handle(self)
return exchange
```

academy/handle.py
```diff
@@ -432,31 +398,35 @@ async def __aexit__(
     def __reduce__(
         self,
     ) -> tuple[
-        type[UnboundRemoteHandle[Any]],
-        tuple[AgentId[AgentT]],
+        type[RemoteHandle[Any]],
+        tuple[AgentId[AgentT], None],
     ]:
-        return (UnboundRemoteHandle, (self.agent_id,))
+        return (RemoteHandle, (self.agent_id, None))
```
gpauloski: If we change the default of `exchange_client` to `None` then we don't need to provide `None` explicitly here.

tests/unit/runtime_test.py
```diff
@@ -548,7 +545,7 @@ async def _request_handler(_: Any) -> None:  # pragma: no cover
 class HandleBindingAgent(Agent):
     def __init__(
         self,
-        unbound: UnboundRemoteHandle[EmptyAgent],
+        unbound: RemoteHandle[EmptyAgent],
```
gpauloski: The name of this test might be more clear as something like `test_runtime_handle_exchange_client_context` and the test agent definition `HandleContextAgent`.

academy/manager.py
AK2000: Yeah, I consider this case in the tests. Basically, the scenario where this is an issue is (1) Neither the exchange client or the manager is used as a context manager and (2) the handle is manually constructed instead of using either the manager or exchange `get_handle` function. In this case we throw an error, and if we rename the error as above, it should be pretty explanatory.

academy/handle.py
AK2000: No, because we want it to re-bind if the context changes. This is the case where you are passing a handle to agents, and the handle does not get serialized. I still need to think about how we want this to work. The scenario where you pass the same handle to 2 agents, or use the handle from both the client and the agent is still problematic. We basically need to force a clone operation in those cases.

academy/manager.py
gpauloski: Make sense. We should probably document all the ways the exchange client context can get set somewhere.

academy/handle.py
gpauloski: To make sure I understand correctly, you only set `self._exchange` if it was already set (i.e., someone specified it in the constructor) and the exchange client in the current context has changed?

gpauloski:
Removing the `client_id` attribute is okay. Doesn't look like it's used outside of some unit tests, and you can still get the `client_id` for the current context with `self.exchange.client_id`.

I'm okay with a separate PR. I'll approve the changes for now and let you decide if you want to do a separate PR.

gpauloski (approved):

PR #181, opened 2025-07-17
Allow handles to be used by multiple clients

### LINKED ISSUE #169: Shared handle references in `Agent.__init__` do not get rebound

AK2000:
## Summary
Follow up to #177. A RemoteHandle is now almost completely decoupled from the exchange client it uses. Every time a message is sent, the handle reads the exchange from the context. This allows the same handle to be used by multiple clients without making copies. The only state that is modified within the handle is a set of registered exchanges. As this set only grows, it should be conflict free. 

(Re-)Introduces BoundRemoteHandle, which is a remote handle with a fixed exchange client. This is for exceptional use cases where the address to send messages might differ from the address set in the context variable.

## Related Issues
- Closes #169 (finally...hopefully)

## Changes

## Pull Request Checklist

### DIFF
--- a/academy/context.py
+++ b/academy/context.py
@@ -55,7 +55,6 @@ class ActionContext:
         if self._source_handle is None:
             self._source_handle = RemoteHandle(
                 self.source_id,
-                self._exchange_client,
             )
         return self._source_handle
 
--- a/academy/context.py
+++ b/academy/context.py
@@ -53,9 +53,7 @@ class ActionContext:
                 'Cannot create handle to source because it is a user entity.',
             )
         if self._source_handle is None:
-            self._source_handle = RemoteHandle(
-                self.source_id,
-            )
+            self._source_handle = RemoteHandle(self.source_id)
         return self._source_handle
 
     def is_agent_source(self) -> bool:
--- a/academy/exchange/__init__.py
+++ b/academy/exchange/__init__.py
@@ -200,12 +200,6 @@ class ExchangeClient(abc.ABC, Generic[ExchangeTransportT]):
         """Close the transport."""
         ...
 
-    async def _close_handles(self) -> None:
-        """Close all handles created by this client."""
-        for key in tuple(self._handles):
-            handle = self._handles.pop(key)
-            await handle.close(wait_futures=False)
-
     async def discover(
         self,
         agent: type[Agent],
@@ -358,7 +352,6 @@ class AgentExchangeClient(
             if self._closed:
                 return
 
-            await self._close_handles()
             await self._transport.close()
             self._closed = True
             logger.info('Closed exchange client for %s', self.client_id)
@@ -425,7 +418,6 @@ class UserExchangeClient(ExchangeClient[ExchangeTransportT]):
             if self._closed:
                 return
 
-            await self._close_handles()
             await self._transport.terminate(self.client_id)
             logger.info(f'Terminated mailbox for {self.client_id}')
             await self._stop_listener_task()
--- a/academy/exchange/__init__.py
+++ b/academy/exchange/__init__.py
@@ -15,6 +15,7 @@ from typing import Generic
 from typing import Protocol
 from typing import runtime_checkable
 from typing import TypeVar
+from weakref import WeakValueDictionary
 
 if sys.version_info >= (3, 11):  # pragma: >=3.11 cover
     from typing import TypeAlias
@@ -166,7 +167,9 @@ class ExchangeClient(abc.ABC, Generic[ExchangeTransportT]):
         transport: ExchangeTransportT,
     ) -> None:
         self._transport = transport
-        self._handles: dict[uuid.UUID, RemoteHandle[Any]] = {}
+        self._handles: WeakValueDictionary[uuid.UUID, RemoteHandle[Any]] = (
+            WeakValueDictionary()
+        )
         self._close_lock = asyncio.Lock()
         self._closed = False
 
@@ -412,7 +415,7 @@ class UserExchangeClient(ExchangeClient[ExchangeTransportT]):
         """Close the user client.
 
         This terminates the user's mailbox, closes the underlying exchange
-        transport, and closes all handles produced by this client.
+        transport.
         """
         async with self._close_lock:
             if self._closed:
--- a/academy/handle.py
+++ b/academy/handle.py
@@ -13,16 +13,12 @@ from typing import Protocol
 from typing import runtime_checkable
 from typing import TYPE_CHECKING
 from typing import TypeVar
+from weakref import WeakSet
 
 if sys.version_info >= (3, 10):  # pragma: >=3.10 cover
     from typing import ParamSpec
 else:  # pragma: <3.10 cover
     from typing_extensions import ParamSpec
-
-if sys.version_info >= (3, 11):  # pragma: >=3.11 cover
-    pass
-else:  # pragma: <3.11 cover
-    pass
 
 from academy.exception import AgentTerminatedError
 from academy.exception import ExchangeClientNotFoundError
@@ -236,21 +232,45 @@ exchange_context: ContextVar[ExchangeClient[Any]] = ContextVar(
 class RemoteHandle(Generic[AgentT]):
     """Handle to a remote agent.
 
+    By default a remote handle uses the 'exchange_context' ContextVar
+    as the exchange client to send messages. This allows the outgoing
+    mailbox to change depending on the context in which the handle is
+    used. The `exchange` argument is used as the default client when
+    the ContextVar is not set. `ignore_context` will cause the handle
+    to only use the `exchange` argument provided in the initializer no
+    matter the context where the handle is used. This should only be for
+    advanced usage.
+
     Args:
         agent_id: EntityId of the target agent of this handle.
+        exchange: Exchange client to use to send messages.
+        ignore_context: Ignore exchange client set in context.
     """
 
     def __init__(
         self,
         agent_id: AgentId[AgentT],
+        exchange: ExchangeClient[Any] | None = None,
+        ignore_context: bool = False,
     ) -> None:
         self.agent_id = agent_id
-        self._registered_exchanges: set[ExchangeClient[Any]] = set()
+        self._exchange = exchange
+        self._registered_exchanges: WeakSet[ExchangeClient[Any]] = WeakSet()
+        self.ignore_context = ignore_context
+
+        if ignore_context and not exchange:
+            raise ValueError(
+                'Cannot initialize handle with ignore_context=True '
+                'and no explicit exchange.',
+            )
 
         # Unique identifier for each handle object; used to disambiguate
         # messages when multiple handles are bound to the same mailbox.
         self.handle_id = uuid.uuid4()
         self._futures: dict[uuid.UUID, asyncio.Future[Any]] = {}
+
+        if self._exchange is not None:
+            self._register_with_exchange(self._exchange)
 
     @property
     def exchange(self) -> ExchangeClient[Any]:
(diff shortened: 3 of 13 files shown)


academy/context.py
```diff
@@ -55,7 +55,6 @@ def source_handle(self) -> Handle[Any]:
         if self._source_handle is None:
             self._source_handle = RemoteHandle(
                 self.source_id,
-                self._exchange_client,
             )
```
gpauloski: Nit: format this to a single line. I'm surprised ruff didn't do this automatically.

academy/exchange/__init__.py
```diff
@@ -358,7 +352,6 @@ async def close(self) -> None:
             if self._closed:
                 return
 
-            await self._close_handles()
```
gpauloski, line 361: The docstring also needs to be updated to remove the handles bit.

academy/handle.py
```diff
@@ -23,17 +20,13 @@
     from typing_extensions import ParamSpec
 
 if sys.version_info >= (3, 11):  # pragma: >=3.11 cover
-    from typing import Self
+    pass
 else:  # pragma: <3.11 cover
-    from typing_extensions import Self
+    pass
```
gpauloski: This can be removed.

A while ago I opened an issue in ruff to autofix this... maybe one day.

academy/handle.py
```diff
@@ -375,74 +263,37 @@ def exchange(self) -> ExchangeClient[Any]:
...
         return remote_method_call
 
-    def clone(self) -> RemoteHandle[AgentT]:
-        """Create an unbound copy of this handle."""
-        return RemoteHandle(self.agent_id)
-
     async def _process_response(self, response: Message[Response]) -> None:
         future = self._futures.pop(response.tag, None)
-        if future is None or future.cancelled():
+        if future is None or future.cancelled():  # pragma: no cover
```
gpauloski: Should this be a warning now that we don't expose the futures? Or maybe even an error if its now an invariant to not get a response back to the handle that we didn't expect.

academy/handle.py
```diff
@@ -508,13 +336,12 @@ async def action(self, action: str, /, *args: Any, **kwargs: Any) -> R:
                 typically indicates the agent shutdown for another reason
                 (it self terminated or via another handle).
             Exception: Any exception raised by the action.
-            HandleClosedError: If the handle was closed.
         """
-        if self._closed:
-            raise HandleClosedError(self.agent_id, self.client_id)
+        exchange = self.exchange
+        self._register_with_exchange(exchange)
```
gpauloski, line 373: Is there a reason not to have this call in the `self.exchange` property? It seems like these always come in a pair:
```
        exchange = self.exchange
        self._register_with_exchange(exchange)
```

academy/handle.py
```diff
@@ -596,20 +422,92 @@ async def shutdown(self, *, terminate: bool | None = None) -> None:
...
+        """
+        return self._exchange
+
+    def __reduce__(
+        self,
+    ) -> tuple[
+        type[BoundRemoteHandle[Any]],
+        tuple[AgentId[AgentT], ExchangeClient[Any]],
+    ]:
+        return (BoundRemoteHandle, (self.agent_id, self._exchange))
```
gpauloski: I think `ExchangeClient` is not pickleable because the underlying transports are not. (But this might be moot w.r.t my other comment on merging the handle types.)

academy/handle.py
```diff
@@ -596,20 +422,92 @@ async def shutdown(self, *, terminate: bool | None = None) -> None:
...
         await self.exchange.send(request)
         logger.debug(
             'Sent shutdown request from %s to %s',
-            self.client_id,
+            exchange.client_id,
             self.agent_id,
         )
+
+
+class BoundRemoteHandle(RemoteHandle[AgentT]):
```
gpauloski: I'd prefer, if we can, keeping just a single `RemoteHandle` type, which can then be consolidated with the `Handle` protocol in a later PR.

Can we stick with the prior approach where `RemoteHandle` takes an optional `ExchangeClient`? We could add an additional flag like `ignore_context` (or a better name?) that ignores any client set in the current context and forces use of the provided client in `__init__`. In that case, `self._registered_exchanges` would always have size 1.

This would give us three usage modes:
1. No client provided so the handle always uses the one in the current context
2. Client provided but `ignore_context=False` (the default) so the current context client is used if available otherwise the provided client is used.
3. Client provided and `ignore_context=True` so only the provided client is used and the context client is ignored (essentially this `BoundRemoteHandle` behavior).

I don't think this should be too complex. Just an extra if in `__init__` to register the handle if provided and some extra logic in the `exchange` property based on the setting of `ignore_context`. Let me know what you think.

academy/handle.py
```diff
@@ -457,40 +308,17 @@ async def _process_response(self, response: Message[Response]) -> None:
...
-            logger.debug('Cancelling pending futures for %s', self)
-            for future in self._futures:
-                self._futures[future].cancel()
-
-    def closed(self) -> bool:
-        """Check if the handle has been closed."""
-        return self._closed
+        if exchange not in self._registered_exchanges:
+            exchange.register_handle(self)
+            self._registered_exchanges.add(exchange)
```
gpauloski, line 353: This creates a strong reference cycle between the `ExchangeClient` and `Handle`. We can use a `weakref.WeakSet` instead.

I'm wondering if we should use a `weakref.WeakValueDictionary` in the `ExchangeClient` too? In that case, the only things with strong references to a `Handle` would be the `Manager` and any user code. If the user discarded all of their handles then they'd automatically get removed from the `WeakValueDictionary` too.

academy/handle.py
AK2000: The idea here was to prevent registering with exchanges that don't need to know about this handle. Just viewing the "current" exchange client does not necessarily mean that this handle is going to receive messages from that exchange. Only once we send a request on the exchange should we start expecting responses on that exchange.

academy/handle.py
AK2000: Yes, but I need to override this method so it doesn't accidentally serialize using the superclass. Instead of throwing an error here, I preferred to let the serialization proceed and throw an error at the actual class that is not serializable.

academy/handle.py
AK2000: I meant to ask about this. What was the case where we would get unexpected responses? Did it have to do with handle closing?

gpauloski: Looking at the git blame, it used to only be the `future.cancelled()` case that was handled for when a handle is closed. I added the `future is None` check when refactoring the method for an unrelated reason, and my guess is I probably just auto-pilot wrote that pattern.

It should be an now invariant that the key is present in `self._futures`, plus futures aren't exposed anymore/no closing means no cancelling. So I think it's safe to just say that's an `AssertionError` and indicative of some other problem.

academy/handle.py
```diff
@@ -314,55 +218,59 @@ async def shutdown(self, *, terminate: bool | None = None) -> None:
...
-        exchange: Exchange client used for agent communication.
         agent_id: EntityId of the target agent of this handle.
+        exchange: Exchange client to use to send messages.
+        ignore_context: Ignore exchange client set in context.
     """
 
     def __init__(
         self,
         agent_id: AgentId[AgentT],
         exchange: ExchangeClient[Any] | None = None,
```
gpauloski, line 255: Can you make these key-word only arguments so that if we add new arguments in the future we don't need to worry about breaking changes to positional argument ordering?
```
    def __init__(
        self,
        agent_id: AgentId[AgentT],
        *,
        ignore_context: bool = False,
    ) -> None:
```

gpauloski: This will require a change to `__reduce__` which doesn't support keyword arguments

academy/handle.py
```diff
@@ -375,77 +283,53 @@ def exchange(self) -> ExchangeClient[Any]:
...
     def __reduce__(
         self,
     ) -> tuple[
         type[RemoteHandle[Any]],
-        tuple[AgentId[AgentT],],
+        tuple[Any, ...],
     ]:
-        return (RemoteHandle, (self.agent_id,))
+        return (
+            RemoteHandle,
```
gpauloski: Should we have this just return `RemoteHandle, (self,agent_id,)` so a serialized handle always reverts to the default use the current context mode?

If `self._exchange is not None` then pickling fails anyways, and the only way for `self.ignore_context` to be different from the default is for `self._exchange` to not be `None`.

(Also makes it possible to have keyword only arguments as I noted above)

tests/unit/handle_test.py
```diff
@@ -95,42 +76,20 @@ async def test_proxy_handle_agent_shutdown_errors() -> None:
...
-        assert isinstance(reconstructed, RemoteHandle)
-        assert reconstructed.agent_id == handle.agent_id
-        # _exchange in handle is empty
-        assert reconstructed._exchange is None
-        # exchange is returned from context variable
-        assert reconstructed.exchange == exchange_client
-        assert str(reconstructed) == str(handle)
-        assert repr(reconstructed) == repr(handle)
+    handle = RemoteHandle(registration.agent_id)
+    # Note: don't call pickle.dumps here because ThreadExchange
```
gpauloski: If we change `__reduce__` to just return the `agent_id` then this test can call `pickle.dumps/loads` directly.

tests/unit/handle_test.py
```diff
@@ -139,94 +98,70 @@ async def test_agent_remote_handle_context() -> None:
...
-    factory = LocalExchangeFactory()
-    exchange_client = await factory.create_user_client()
-    registration = await exchange_client.register_agent(EmptyAgent)
-    async with RemoteHandle(registration.agent_id, exchange_client) as handle:
-        cloned = handle.clone()
 
-        with pytest.raises(ExchangeClientNotFoundError):
-            assert cloned.exchange == exchange_client
+    exchange_context.set(exchange_client)
+    assert handle.exchange == exchange_client
```
gpauloski: I think when checking the value of `handle.exchange` we should use `is` rather than `==`. Practically it has the same result here because we don't implement `__eq__` on `Handle`, but semantically they have different meanings.

tests/unit/runtime_test.py
```diff
@@ -286,6 +283,36 @@ async def test_runtime_shutdown_message(
...
 
+class InfiniteAgent(Agent):
+    @loop
+    async def wait(self, shutdown: asyncio.Event) -> None:
+        while True:
+            await asyncio.sleep(5)
+
+
+@pytest.mark.asyncio
+async def test_runtime_cancel_loop(
```
gpauloski, line 294: Is the addition of this test because one of the removed tests was incidentally responsible for coverage of loop cancellation?

gpauloski (approved):
Looks great! I think this will be way better than the previous find, bind, and replace logic. Left some comments for a few minor fixes, but I'll go ahead and approve.

tests/unit/runtime_test.py
AK2000: Yes, but it only came up in py3.9

academy/handle.py
AK2000: Good point

PR #184, opened 2025-07-18
Re-export exchange factories and transports

AK2000:
## Summary
Re-export exchange factories and transports from `academy.exchange.__init__.py`. Move `ExchangeClient` and `ExchangeFactory` definitions into their own class.

## Changes

## Pull Request Checklist

### DIFF
--- a/academy/exchange/factory.py
+++ b/academy/exchange/factory.py
@@ -1,17 +1,12 @@
 from __future__ import annotations
 
 import abc
-import asyncio
-import contextlib
 import logging
 import sys
-import uuid
 from collections.abc import Coroutine
-from types import TracebackType
 from typing import Any
 from typing import Callable
 from typing import Generic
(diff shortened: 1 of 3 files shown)


academy/exchange/factory.py
```diff
@@ -0,0 +1,452 @@
...
+        user_id = transport.mailbox_id
+        assert isinstance(user_id, UserId)
+        return UserExchangeClient(
+            user_id,
+            transport,
+            start_listener=start_listener,
+        )
+
+
+class ExchangeClient(abc.ABC, Generic[ExchangeTransportT]):
```
gpauloski: It might make more sense to have these in their own `academy.exchange.client` module?

gpauloski (dismissed):
I think `factory.py` could be split into `factory.py` and `client.py` but otherwise looks good.

PR #190, opened 2025-07-21
Auth ID Cacheing

AK2000:
## Summary
Added cache and local client to import auth performance. Previously, a `oauth2_token_introspect` call was made for every API call to the cloud hosted server. In addition, the AuthClient was protected by a lock, so only one call could proceed even with asyncio. To fix these issues, I moved the AuthClient to thread local storage and included a cache on the token introspect calls. Preliminary benchmarking on my (very old) laptop shows that this led to a significant speed-up of authentication compared to before. There is more work to be done, but this likely explains most of the performance issues we were having during the tutorial.

Previous benchmark (the variance is likely due to the range of other tasks on my laptop, and the fact that I only ran 1 repetition):
```
Clients: 1      Messages: 100   Time: 33.190674054      Throughput: 3.012894520831475
Clients: 2      Messages: 100   Time: 139.076211315     Throughput: 1.4380604569893767
Clients: 3      Messages: 100   Time: 184.203588142     Throughput: 1.6286327700019294
Clients: 4      Messages: 100   Time: 236.53706735      Throughput: 1.6910668779372604
```

After P.R.
```
Clients: 1      Messages: 100   Time: 0.514027066       Throughput: 194.54228505547215
Clients: 2      Messages: 100   Time: 1.701158491       Throughput: 117.56694103347952
Clients: 3      Messages: 100   Time: 1.406130397       Throughput: 213.35147909472295
Clients: 4      Messages: 100   Time: 1.817344193       Throughput: 220.10139936106202
```

And without auth:
```
Clients: 1      Messages: 100   Time: 0.32993737        Throughput: 303.087825425777
Clients: 2      Messages: 100   Time: 0.880020703       Throughput: 227.26738054934145
Clients: 3      Messages: 100   Time: 1.464129131       Throughput: 204.8999597426902
Clients: 4      Messages: 100   Time: 1.990686387       Throughput: 200.93571876120936
```

## Changes

## Pull Request Checklist

### DIFF
--- a/academy/exchange/cloud/authenticate.py
+++ b/academy/exchange/cloud/authenticate.py
@@ -2,13 +2,17 @@
 
 from __future__ import annotations
 
+import asyncio
+import threading
 import uuid
 from collections.abc import Mapping
-from threading import Lock
+from concurrent.futures import ThreadPoolExecutor
 from typing import Protocol
 from typing import runtime_checkable
 
 import globus_sdk
+from cachetools import cached
+from cachetools import TTLCache
 
 from academy.exchange.cloud.config import ExchangeAuthConfig
 from academy.exchange.cloud.exceptions import ForbiddenError
@@ -20,7 +24,7 @@ from academy.exchange.cloud.login import AcademyExchangeScopes
 class Authenticator(Protocol):
     """Authenticate users from request headers."""
 
-    def authenticate_user(self, headers: Mapping[str, str]) -> uuid.UUID:
+    async def authenticate_user(self, headers: Mapping[str, str]) -> uuid.UUID:
         """Authenticate user from request headers.
 
         Warning:
@@ -43,7 +47,7 @@ class Authenticator(Protocol):
 class NullAuthenticator:
     """Authenticator that implements no authentication."""
 
-    def authenticate_user(self, headers: Mapping[str, str]) -> uuid.UUID:
+    async def authenticate_user(self, headers: Mapping[str, str]) -> uuid.UUID:
         """Authenticate user from request headers.
 
         Args:
@@ -69,9 +73,10 @@ class GlobusAuthenticator:
         audience: Intended audience of the token. This should typically be
             the resource server of the the token was issued for. E.g.,
             the UUID of the ProxyStore Relay Server application.
-        auth_client: Optional confidential application authentication client
-            which is used for introspecting client tokens.
     """
+
+    TOKEN_CACHE_LIMIT: int = 1024
+    TOKEN_TTL: int = 60
 
     def __init__(
         self,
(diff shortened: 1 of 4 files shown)


academy/exchange/cloud/authenticate.py
```diff
@@ -69,30 +73,51 @@ class GlobusAuthenticator:
...
 
-    def authenticate_user(self, headers: Mapping[str, str]) -> uuid.UUID:
+    @property
+    def auth_client(self) -> globus_sdk.ConfidentialAppAuthClient:
+        """A thread local copy of the Globus AuthClient."""
+        try:
+            return self._local_data.auth_client
+        except AttributeError:
+            self._local_data.auth_client = (
+                globus_sdk.ConfidentialAppAuthClient(
```
yadudoc, line 109: I skimmed the `ConfidentialAppAuthClient` and did not find any troublesome non-thread-safe bits that could interfere with thread-local stores, which would require locks in the cache.

academy/exchange/cloud/authenticate.py
```diff
@@ -69,30 +73,51 @@ class GlobusAuthenticator:
         audience: Intended audience of the token. This should typically be
             the resource server of the the token was issued for. E.g.,
             the UUID of the ProxyStore Relay Server application.
-        auth_client: Optional confidential application authentication client
-            which is used for introspecting client tokens.
     """
 
+    TOKEN_CACHE_LIMIT: int = 1024
+    TOKEN_TTL: int = 60
```
gpauloski: Can you append the unit to this var? I.e., `TOKEN_TTL_SECONDS`

academy/exchange/cloud/authenticate.py
```diff
@@ -69,30 +73,51 @@ class GlobusAuthenticator:
...
-            globus_sdk.ConfidentialAppAuthClient(
-                client_id=str(client_id),
-                client_secret=str(client_secret),
-            )
-            if auth_client is None
-            else auth_client
-        )
-        self.auth_client_lock = Lock()
+        self._local_data = threading.local()
+        self.executor = ThreadPoolExecutor()
```
gpauloski: Can you set the `thread_name_prefix` on this? It helps when debugging to have the threads named.

academy/exchange/cloud/authenticate.py
```diff
@@ -69,30 +73,51 @@ class GlobusAuthenticator:
...
+        except AttributeError:
+            self._local_data.auth_client = (
+                globus_sdk.ConfidentialAppAuthClient(
+                    client_id=str(self.client_id),
+                    client_secret=str(self.client_secret),
+                )
+            )
+            return self._local_data.auth_client
+
+    @cached(
```
gpauloski: This might need to be `cachedmethod` otherwise a reference to `self` gets cached preventing the object from being garbage collected (at least that's how the stdlib LRU cache is).

academy/exchange/cloud/authenticate.py
gpauloski: I guess there is some tradeoff here between performance and guaranteeing the validity of the token. 60 seconds seems ok to me.

gpauloski (changes requested):
Left a couple small requests but otherwise looks good. Nice performance improvements!

academy/exchange/cloud/authenticate.py
yadudoc: The defaults are from `globus-compute`'s hosted service. Longer TTLs mess with token invalidation, and we can revisit when we know our workload better.

academy/exchange/cloud/authenticate.py
yadudoc: Nice catch!

yadudoc (dismissed):
Looks good to me. I've posted some minor comments and some responses to @gpauloski.

academy/exchange/cloud/authenticate.py
AK2000: I don't know that it matters in this case, but I thought the point of the thread local store was that it does not need to be thread-safe because each instance will only be used by one thread.

academy/exchange/cloud/authenticate.py
AK2000: Ope, that's actually much better because we can make them configurable now

academy/exchange/cloud/authenticate.py
yadudoc: True, but that still requires `ConfidentialAppAuthClient` to play nice and not rely on any global singleton business for example. Either way, this is safe, no locks required.

PR #193, opened 2025-07-24
Create storage protocol for HTTP exchange with local and Redis implementations

AK2000:
## Summary
We talked about supporting a Redis backend to the exchange server so we can have multiple workers serving requests to the same mailboxes. We still need to support a local http exchange to work with local processes (or even local cluster deployments).
To support both Redis and the local exchange, I've refactored the mailbox manager into the MailboxBackend protocol, and added the RedisBackend as an implementation of the protocol.

Part of implementing a multi-user exchange (#82)

## Changes

## Pull Request Checklist

### DIFF
--- a/academy/exception.py
+++ b/academy/exception.py
@@ -61,7 +61,13 @@ class MessageTooLargeError(ExchangeError):
     class an way to by-pass the exchange for large data.
     """
 
-    pass
+    def __init__(self, size: int, limit: int) -> None:
+        self.size = size
+        self.limit = limit
+
+        super().__init__(
+            f'Message of size {size} bytes is larger than limit {limit}.',
+        )
 
 
 class MailboxTerminatedError(ExchangeError):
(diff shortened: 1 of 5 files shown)


academy/exception.py
```diff
@@ -53,6 +53,17 @@ class ForbiddenError(ExchangeError):
...
 
+class MessageTooLargeError(ExchangeError):
+    """Message payload is too large for exchange.
+
+    If encountering this error, consider using the
+    [`ProxyStoreExchangeTransport`][academy.exchange.ProxyStoreExchangeTransport]
+    class an way to by-pass the exchange for large data.
+    """
+
+    pass
```
yadudoc: It's useful for this error to tell the user both the size of the offending message and the limit.

academy/exchange/cloud/config.py
```diff
@@ -50,6 +55,55 @@ class because they often contain secrets.
...
+        hostname: Redis host
+        port: Redis port
+        kwargs: Any additional args to Redis
+    """
+
+    model_config = ConfigDict(extra='forbid')
+
+    hostname: str = 'localhost'
+    port: int = 6379
+    message_size_limit_kb: int = 1024
```
yadudoc: We should limit this to redis' limit of 512MB

PR #214, opened 2025-09-23
Support for passing kwargs to submit at agent launch

### LINKED ISSUE #211: Passing Resource Specification to Executor
### Describe the problem.

Some Parsl [executors](https://parsl.readthedocs.io/en/stable/userguide/configuration/executors/wqex.html#heterogeneous-resources) expect a resource specification to be passed as part of submitting a task. Currently, there is no support for doing so within the manager launch command.

```console
github
```

### Version

master

### Python Version

3.12

### OS and Platform

x86 Linus

AK2000:
## Summary
Adds an additional argument to the `manager.launch` interface to allow the passing of key word arguments to the `executor.submit`. This interface is meant to support things like the `parsl_resource_specification` when launching agents so that Academy can be used with the WorkQueue executor.

## Related Issues
- Closes #211

## Changes

## Pull Request Checklist

### DIFF
--- a/tests/unit/manager_test.py
+++ b/tests/unit/manager_test.py
@@ -3,11 +3,16 @@ from __future__ import annotations
 import asyncio
 import os
 import pathlib
+import sys
 from concurrent.futures import Future
 from concurrent.futures import ThreadPoolExecutor
 from typing import Any
 from typing import Callable
-from typing import ParamSpec
+
+if sys.version_info >= (3, 10):  # pragma: >=3.10 cover
+    from typing import ParamSpec
+else:  # pragma: <3.10 cover
+    from typing_extensions import ParamSpec
 
 import pytest
 


academy/manager.py
```diff
@@ -45,6 +45,7 @@ class _RunSpec(Generic[AgentT, ExchangeTransportT]):
     registration: AgentRegistration[AgentT]
     agent_args: tuple[Any, ...]
     agent_kwargs: dict[str, Any]
+    submit_kwargs: dict[str, Any]
```
benclifford, line 49: so there's an option to supply your own custom kwargs but not your own custom positional args?

AK2000: Hmm, I guess I don't see a case where an executor parses positional-only arguments for runtime parameters --- I think this wouldn't make sense as a design choice.

academy/manager.py
```diff
@@ -73,6 +74,7 @@ async def _run_agent_on_worker_async(
 
 def _run_agent_on_worker(
     spec: _RunSpec[AgentT, ExchangeTransportT],
+    **kwargs: Any,
```
benclifford, line 89: matching any/all args and then ignoring them is always a bit suspicious usability-wise: it suppresses a bunch of type-driven error checking such as a user requesting "magic_feature=True" and then wondering why the magic doesn't happen, rather than being told it's unimplemented.

AK2000: How would this be dealt with in Parsl --- the specific use case is allowing the resource specification of the WorkQueueExecutor. It seems like we need to accept `parsl_resource_specification` like [here](https://parsl.readthedocs.io/en/stable/userguide/configuration/executors/wqex.html#heterogeneous-resources), and I wanted a way to do that without having to explicitly tailor the method.

academy/manager.py
benclifford: right this is *very* parsl specific, where you're still trying to pretend parsl has a concurrent.futures.Executor API on the front, which it doesn't really...

academy/manager.py
benclifford: In Parsl-land the things you are running know they are running inside Parsl and know what to expect because of that.

I guess the decision here is do you want to specialise on knowing you're potentially running inside Parsl, or do you want to support a more abstract "any kwargs, for anything you can imagine - even though we only know of one"

academy/manager.py
AK2000: Dask has something [similar](https://distributed.dask.org/en/stable/resources.html#resources-are-applied-separately-to-each-worker-process). But I don't think the `resources` argument is forwarded to the executed function.

academy/manager.py
AK2000: Sorry I didn't come base to this for a while. I think my preference at this point would be to leave it more general. I don't think type-checking on this specific method will affect usability that much because it is internal.

Issue #232, opened 2025-10-22
RedisExchangeTransport.discover makes terrible confusions between `str` and `bytes` returned from `redis` API in SCANs and key values


benclifford:
### Describe the problem.

The code assumes that the redis client returns strings for Redis `SCAN`s using `scan_iter` and in retrieving the key value for state later in the function.

https://github.com/academy-agents/academy/blob/a1d8864353ed6851503bb296b12a0d86b87af9a1/academy/exchange/redis.py#L143

The assumption is not just in that assertion but in all the mechanics around string values throughout that function. I made a messy hack to make progress on my project and see that `discover` does otherwise work, but that's not something any other human may see.

This code has test coverage, but not against Redis, instead falling victim to mocking.

Here's a cut down reproduction of the misassumed functionality. You'll need a real redis, with some real keys in it - my academy runs have left enough in my install just through leaving stuff behind:

```
>>> import redis
>>> r=redis.Redis()
>>> type(next(r.scan_iter("*")))
<class 'bytes'>
```

```
$ redis-server  --version
Redis server v=8.0.2 sha=00000000:0 malloc=jemalloc-5.3.0 bits=64 build=5a33a2b1553c5bfc
```

```
$ pip show redis
Name: redis
Version: 6.4.0
Summary: Python client for Redis database and key-value store
Home-page: https://github.com/redis/redis-py
Author: 
Author-email: "Redis Inc." <oss@redis.com>
License-Expression: MIT
Location: /home/benc/parsl/academy/virtualenv-trixie/lib/python3.13/site-packages
Requires: 
Required-by: academy-py
```

### How did you install and run the code?

```console
pip install -e .   # from my hacked up repo
```

### Version

71f8ef1a1f4650c3edf1a9aaadc9340e22a90415 + local hacks

### Python Version

3.13

### OS and Platform

x86 linux debian trixie

AK2000:
I think this issue raises two action items that we need to address:

- [ ] Fix the mechanics around string vs. byte data types in RedisExchange and in MockRedis
- [ ] Set up (integration?) testing against Redis in a container, possibly as a Github action

gpauloski:
[...]

PR #233, opened 2025-10-22
fix redis exchange discover

danielrosendo:
## Summary

Redis get() returns bytes, but the code expected a str (e.g., assert isinstance(mro_str, str)).
Adding .decode("utf-8") restores the intended behavior (working with text, not bytes).

## Related Issues

- Closes #XX
- Related to #XX _(if applicable)_

## Changes

- [X] Bug fix (non-breaking change which fixes an issue)

## Testing

N/A

## Pull Request Checklist

### DIFF
--- a/academy/exchange/cloud/backend.py
+++ b/academy/exchange/cloud/backend.py
@@ -621,19 +621,16 @@ class RedisBackend:
             allow_subclasses: Include agents that inherit from the target.
         """
         found: list[AgentId[Any]] = []
-        fqp = agent.encode()
         async for key in self._client.scan_iter(
             'agent:*',
         ):  # pragma: no branch
-            mro_str = await self._client.get(key)
-            assert isinstance(mro_str, bytes), (
-                f'mro_str is {type(mro_str)} with repr {mro_str!r}'
-            )
-            mro = mro_str.split(b',')
-            if fqp == mro[0] or (allow_subclasses and fqp in mro):
-                k = key.split(b':')[-1]
-                sk = k.decode()
-                aid: AgentId[Any] = AgentId(uid=uuid.UUID(sk))
+            mro_str = (await self._client.get(key)).decode()
+            assert isinstance(mro_str, str)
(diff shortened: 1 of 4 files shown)


AK2000:
Haha, I'm not sure if this was best development practice or not, but I took the opportunity to address many issues related to the same thing that were not breaking external functionality but were still leading to things not working as expected.

academy/exchange/hybrid.py
```diff
@@ -225,22 +225,30 @@ async def discover(
         allow_subclasses: bool = True,
     ) -> tuple[AgentId[Any], ...]:
         found: list[AgentId[Any]] = []
-        fqp = f'{agent.__module__}.{agent.__name__}'
+        fqp_str = f'{agent.__module__}.{agent.__name__}'
+        fqp = fqp_str.encode()
```
benclifford: same comment as for discover about live in `str` land and `decode` immediately after `.get`

academy/exchange/cloud/backend.py
```diff
@@ -621,21 +621,28 @@ async def discover(
...
             'agent:*',
         ):  # pragma: no branch
             mro_str = await self._client.get(key)
-            assert isinstance(mro_str, str)
-            mro = mro_str.split(',')
-            if agent == mro[0] or (allow_subclasses and agent in mro):
-                aid: AgentId[Any] = AgentId(uid=uuid.UUID(key.split(':')[-1]))
+            assert isinstance(mro_str, bytes), (
+                f'mro_str is {type(mro_str)} with repr {mro_str!r}'
+            )
```
benclifford: i think this code is a mess between `bytes` and `str` -- I think there's no reason to be working with both types. I think that it would be cleaner to convert everything to `str` at the point that `_client.get` is invoked -- like in the rest of this PR.

Then everything else in this code stays the same, in `str`-land.

(this code is lifted from a hack patch I pasted on slack which I deliberately didn't include in my issue #232)

PR #239, opened 2025-10-27
Support Manager with Agents

AK2000:
## Summary
The `Manager` class is the most convenient for launching new agents, but it is bound to the UserExchangeClient type. There are several use cases for a manager to be launched within an Agent that shouldn't necessitate creating a new mailbox (see @benclifford's response in #236). 

This PR changes the type annotation on the UserExchangeClient, and adds a option in the `Manager.close` method to leave the ExchangeClient open. These changes allow a `Manager` to easily be used within an agent. 

## Changes
- [ ] Breaking (backwards incompatible changes to public interfaces)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] Enhancement (non-breaking change or feature addition)
- [ ] Refactor (internal code or design clean up)
- [ ] Documentation (no changes to the code)
- [ ] Test (changes or additions to testing)
- [ ] Build (change to CI workflows or build processes)
- [ ] Package (changes to package metadata or dependency versions)

## Pull Request Checklist

Please confirm the PR meets the following requirements.
- [x] Relevant tags are added based on the types of changes.
- [x] Code changes pass `pre-commit` (e.g., ruff, mypy, etc.).
- [x] Tests have been added to show the fix is effective or that the new feature works.
- [x] New and existing unit tests pass locally with the changes.
- [x] Docs have been updated and reviewed if relevant.

Issue #236, opened 2025-10-24
Allow action to be asynchronous iterators


WardLT:
### Describe the Request

I would like an action to provide results incrementally rather than waiting to return all at once.

### Sample Code

```python
class SlowLoopAction(Agent):

   @action
   async def make_stuff_slowly(self):
      for i in range(3):
         yield i

agent = SlowLoopAction()
async for r in agent.make_stuff_slowly():
   r
```

PR #240, opened 2025-10-27
Change Manager to Launch Agents in 'event_loop'

AK2000:
## Summary
With the move to asyncio, agents need not be spawned in another thread or process from the script. This PR introduces the ability to run an agent in the current asyncio event loop. This was suggested by @benclifford in PR #239 as a mechanism needed when launching agents within agents.

This PR woroks by allowing an executor in the `Manager._executors` attribute to be None. This does not change the default behavior of the Manager.launch or any other method --- you must explicitly pass `executors=None` or pass a dictionary with `{name: None}` to get this behavior.

## Changes

## Pull Request Checklist

### DIFF
--- a/academy/manager.py
+++ b/academy/manager.py
@@ -53,7 +53,7 @@ class _RunSpec(Generic[AgentT, ExchangeTransportT]):
     logfile: str | None = None
 
 
-async def _run_agent_on_worker_async(
+async def _run_agent_async(
     spec: _RunSpec[AgentT, ExchangeTransportT],
 ) -> None:
     try:
@@ -89,7 +89,7 @@ def _run_agent_on_worker(
 
         init_logging(level=spec.loglevel, logfile=logfile)
 
-    asyncio.run(_run_agent_on_worker_async(spec))
+    asyncio.run(_run_agent_async(spec))
 
 
 @dataclasses.dataclass
@@ -365,7 +365,7 @@ class Manager(Generic[ExchangeTransportT], NoPickleMixin):
                         loop=loop,
                     )
                 else:
-                    await _run_agent_on_worker_async(spec)
+                    await _run_agent_async(spec)
             except asyncio.CancelledError:  # pragma: no cover
                 logger.warning('Cancelled %s task', agent_id)
                 raise
@@ -667,9 +667,7 @@ class Manager(Generic[ExchangeTransportT], NoPickleMixin):
             )
 
 
-def _infer_max_workers(
-    executor: Executor,
-) -> int | None:  # pragma: no cover
+def _infer_max_workers(executor: Executor) -> int | None:  # pragma: no cover
     """Infer the maximum workers of the executor.
 
     The [`Executor`][concurrent.futures.Executor] specification does not


academy/manager.py
```diff
@@ -649,12 +661,17 @@ def _warn_executor_overloaded(
...
+def _infer_max_workers(
+    executor: Executor | None,
+) -> int | None:  # pragma: no cover
     """Infer the maximum workers of the executor.
 
     The [`Executor`][concurrent.futures.Executor] specification does not
     provide a standard mechanism to get the maximum number of workers.
     """
+    if executor is None:
+        return None
```
gpauloski: I think it's better to keep `_infer_max_workers` and `_warn_executor_overloaded` the same and check if `executor is not None` before calling `_warn_executor_overloaded`. In other words, don't bury the edge case check at the bottom of the call stack.

academy/manager.py
```diff
@@ -127,7 +127,9 @@ class Manager(Generic[ExchangeTransportT], NoPickleMixin):
         executors: An executor instance or mapping of names to executors to
             use to run agents. If a single executor is provided, it is set
             as the default executor with name `'default'`, overriding any
-            value of `default_executor`.
+            value of `default_executor`. Setting `executor=None` or passing
+            `{'name': None, ...}` can be used to launch agents in the current
+            event loop.
```
gpauloski: The more I think about this I wonder if running agents in the event loop should be the default behavior and we have the signature as follows?

```python
        executors: MutableMapping[str, Executor] | None = None,
        *,
        default_executor: str = "event_loop",
```
The `__init__` logic gets way simpler.
```python
# Maybe also check that "event_loop" is not in executors
self.executors: dict[str, Executor | None] = {'event_loop': None}
if executors is not None:
    self.executors.update(executors)
```

This does end up being a breaking change because of `default_executor`, but it feels more natural to me. If I want to test out a couple agents locally, why create a thread pool when I already have a perfectly good event loop. 

To recover the old behavior of this:
```python
) as manager:
    # Runs in ThreadPoolExecutor by default
    agent_handle = await manager.launch(ExampleAgent())
```
You'd have to do this (rather than pass the executor name to every call to `launch()`):
```python
    executors={'thread': ThreadPoolExecutor()},
    default_executor='thread',
) as manager:
    # Runs in ThreadPoolExecutor by default
    agent_handle = await manager.launch(ExampleAgent())
```
It's slightly more verbose, but I do think it's more explicit about what's happening.

If we make this change we'll also want to update some of the verbiage around the `Manager` to frame it as "manages agents running locally or within executors" (some docs refer to `Manager` and `Executor` being tightly coupled things.

academy/manager.py
```diff
@@ -150,12 +152,18 @@ def __init__(
             executors = {'default': executors}
             default_executor = 'default'
 
+        if executors is None:
+            executors = {'native': None}
+            default_executor = 'native'
```
gpauloski: I think "event_loop" is more clear than "native".

academy/manager.py
```diff
@@ -347,14 +356,17 @@ async def _run_agent_in_executor(
...
+                    await asyncio.wrap_future(
+                        executor.submit(
+                            _run_agent_on_worker,  # type: ignore[arg-type]
+                            spec,
+                            **spec.submit_kwargs,
+                        ),
+                        loop=loop,
+                    )
+                else:
+                    await _run_agent_on_worker_async(spec)
```
gpauloski: Do we still want the setup logic in `_run_agent_on_worker_async` to be called? I know it's opt-in and not the default, but I can see how someone would look at the signature, see `init_logging` is `False` by default, and think "of course I want my agent to have logging" even though it's in the main event loop process with logging already enabled.

academy/manager.py
```diff
@@ -449,7 +461,7 @@ async def launch(  # noqa: PLR0913
         )
 
         task = asyncio.create_task(
-            self._run_agent_in_executor(executor_instance, spec),
+            self._run_agent(executor_instance, spec),
             name=f'manager-run-{agent_id}',
```
gpauloski: For debugging, it may be handy to set the name of the asyncio task to include the name of the executor it's running in. That way if you dump the running tasks and see a lists of managed agents you know which are where.

academy/manager.py
benclifford: > if running agents in the event loop should be the default behavior 

coming to this codebase well after it switched to asyncio, the api around executors at they are now seems a bit weird to me. The mental model I kinda converged on is:

* agents run in an "event loop+decorations (eg connection to exchange)"
* multiple agents can run in the same "event loop+decorations", not just the same Python process  (that's what this weeks work is about, around @AK2000's PRs and my iterators prototype)
* it *does* make sense to be conscious of where your agent runs: a "mostly idle" router/orchestrator agent can probably run happily alongside many other agents in the same event loop. and if its long running it probably *should* run alongside other stuff. vs  a processing-intense one might want to be running somewhere else and with its own dedicated event-loop thread, with the user conscious that this is blocking an entire worker for the lifetime of the agent so is heavyweight in that sense too. so that stuff *should* be exposed to the user somehow.

benclifford:
I tried this out with my iterator prototype from #236, as I did with #239. All I changed was switching `Manager(executors=...)` to `executors=None`. It runs ok. I can see from logging that the 2nd remote agent (the generator-proxy-agent) now runs in the same thread (the main thread of the remote process) as the 1st remote agent, which is what I would expect if they're all now in the same event loop.

academy/manager.py
benclifford: I think this is already causing problems before this PR, with the thread pool executor -- logging is scoped per process, and thread pools run multiple remote agents in the same process. At least I noticed duplicate log lines that I think are explained by multiple log initialisations in the same process this way.

benclifford: (over in Parsl land, it's been a confused story for many years about what is being initialized for the hosting process vs what is being initialised for a DFK -- especially around logging)

academy/manager.py
benclifford: informally i've been using the term "adjacent" to myself. or "here". so many choices.

academy/manager.py
benclifford: ... and API-wise I kinda want `whatever_handle = self.run_alongside(new_agent)` where self is the self of an agent inside an `@action` and none of this manager stuff exposed.

AK2000: Hmm, I still like that passing a single executor (i.e. instead of a mapping) becomes the default. I think this is more intuitive (if I want to use an executor, I can just pass that executor without having to give it a name and then set it as the default) and avoids a breaking change. I'm on board with having None be the default value.

AK2000: I agree that manually creating the Manager is a bit cumbersome and relies on people to do know that they have to clean up the manager. I moved this to a new PR (#242)

academy/manager.py
AK2000: I went with "event_loop" 🤷

academy/manager.py
gpauloski: > the api around executors at they are now seems a bit weird

For context, the APIs around executors exist to solve the problem of "what's the easiest way to run an agent in an event loop on a remote process?"

I do think it's odd that our examples show running these async things (agents) in different threads of the main process---kinda defeats the point. This is mostly a vestige of the pre-async days. In addition, a lot of the decisions in the current API were made around the assumption that most people will run their agents on remote resources and the user's python script is just this lightweight thing to spawn all of them. That's why the `Manager` is designed around executors.

For this PR I initially considered proposing a new interface for running agents in the current event loop and keeping the `Manager` as the thing that's tightly coupled with running agents in executors. But the `Manager` is currently the main entrypoint for new users so it seems worth it to expand the scope of the `Manager` to support local execution of agents.

> ... and API-wise I kinda want whatever_handle = self.run_alongside(new_agent) where self is the self of an agent inside an @action and none of this manager stuff exposed.

I tend to agree with this, but I'll move this discussion about implementation details to #242. I worry about having hard-to-reason about nested managers and agents.

academy/manager.py
```diff
@@ -127,7 +128,9 @@ class Manager(Generic[ExchangeTransportT], NoPickleMixin):
         executors: An executor instance or mapping of names to executors to
             use to run agents. If a single executor is provided, it is set
             as the default executor with name `'default'`, overriding any
-            value of `default_executor`.
+            value of `default_executor`. Setting `executor=None` (the default)
+            or passing `{'name': None, ...}` can be used to launch agents
```
gpauloski, line 132: Is there a reason to provide `{'name': None, ...}` if we say `'event_loop'` is always a valid option?

academy/manager.py
```diff
@@ -347,14 +355,17 @@ async def _run_agent_in_executor(
...
+                    await asyncio.wrap_future(
+                        executor.submit(
+                            _run_agent_on_worker,  # type: ignore[arg-type]
+                            spec,
+                            **spec.submit_kwargs,
+                        ),
+                        loop=loop,
+                    )
+                else:
+                    await _run_agent_on_worker_async(spec)
```
gpauloski: With the change to where logging is initialized this could just be called `_run_agent_async`

academy/manager.py
```diff
@@ -649,7 +667,9 @@ def _warn_executor_overloaded(
             )
 
 
-def _infer_max_workers(executor: Executor) -> int | None:  # pragma: no cover
+def _infer_max_workers(
+    executor: Executor,
+) -> int | None:  # pragma: no cover
```
gpauloski: Nit: collapse the lines to reduce code churn

academy/manager.py
AK2000: I can imagine use cases (i.e. the dictionary is from agent names to executors, and you want some agent to run locally or when testing, you switch something that will be a `ProcessPoolExecutor` in production to `None` in testing). Since putting this constraint gives us nothing, I don't want to include it.

PR #242, opened 2025-10-28
Add Manager to Agent on_startup

AK2000:
## Summary
The motivation was that we want a canonical way for Agents to launch other Agents. The manager class provides all of the nice utilities for running agents (i.e. retries, clean up, etc.). So, we want Agents to have a manager to use to launch children.

After experimenting with several options, I added a method to the `Agent` base class called `_agent_startup` that is called by the runtime, and initialize the `Manager` there. Since the Manager uses so many other classes, essentially all references to the `acdemy.agent` module needed to be guarded by `if TYPE_CHECKING:`. Let me know if there is a better way to avoid a circular import.

Other implementation ideas:

- Adding a `Manager` to the `AgentContext`. This would be ideal because the Manager seems well coupled with other parts of the context (i.e. the `ExchangeClient`). The problem is that the context is constructed by the `Runtime`. Then the `Manager` uses the `Runtime` to launch agents, and the `Runtime` constructs a `Manager` to run the agents, introducing an import dependency --- one that's hard to get around without some significant refactoring.
- Constructing the Manager in the existing `agent_on_startup` method. I didn't like this because we instruct people to overwrite this method in their agent implementations, which means they would have to remember to call `super().agent_on_startup()` in order for the manager to be constructed properly. 

## Changes

## Testing

N/A

## Pull Request Checklist

### DIFF
--- a/academy/agent.py
+++ b/academy/agent.py
@@ -761,17 +761,11 @@ class Agent:
         args: tuple[Any, ...] | None = None,
         kwargs: dict[str, Any] | None = None,
     ) -> Handle[AgentT]:
-        """Launch an child agent in the current event loop.
-
-        Note:
-           This is a wrapper around `self.manager.launch`. For more options
-           and control over launching child agents, use `self.manager`
-           directly.
+        """Launch a child agent in the current event loop.
 
         Args:
-            agent: Agent instance the agent will implement or the
-                agent type that will be initialized on the worker using
-                `args` and `kwargs`.
+            agent: An agent instance to launch, or an agent class to be
+                instantiated before launching using `args` and `kwargs`.
             args: Positional arguments used to initialize the agent.
                 Ignored if `agent` is already an instance.
             kwargs: Keyword arguments used to initialize the agent.
--- a/tox.ini
+++ b/tox.ini
@@ -21,7 +21,7 @@ commands =
 extras = dev
 commands =
     coverage erase
-    coverage run -m pytest -s tests/integration {posargs}
+    coverage run -m pytest tests/integration {posargs}
     coverage combine --quiet
     coverage report --include tests/integration/*
 


gpauloski:
Gonna skip the details for now to deliberate on higher-level design questions.

I wonder if this is the best solution to the problem of spawning child agents, or if it's just the easiest path to implementing `agent_launch_alongside`?

There are different versions to launching a child agent:
1. Is the child running in the same event loop as the parent or in a different process?
2. Is the child's scope/lifetime limited to the parent's scope, or is the child detached from the parent (i.e., may outlive the parent)?

Which of those are we wanting to support? 

This solution makes `same event loop`+`same scope` easy, but it doesn't make `different event loop`+`same scope` much easier. You still have to muck around with creating an executor, adding it to the manager, and calling `Manager.launch()`. You save constructing the manager and closing it but that's pretty trivial.

We don't have any solution at the moment for creating a child agent that would outlive its parent, so that's a whole other conversation. But the `Manager` is pretty clear about the lifetimes of child agents being associated with the lifetime of the `Manager`, so we should keep that in mind.

Does it make sense to have many managers in the same event loop/process? If I write a script that uses a manager to spawns agent A (in the same event loop) and then A spawns a child B (again, in the same event loop using `agent_launch_alongside`), we now have three different managers (client, A, and B) for three agents in one event loop. This seems like a code smell to me and makes any introspection/debugging much more complicated to do and reason about. For instance, `Manager.running()` on the top-level manager only returns `A`, not `B`, even though the top-level manager does "own" `B` implicitly through `A`.

This is still a problem if you write your own agent that uses a `Manager`, but I think it's worth thinking more about what the intended use cases are for these interfaces and what contracts we want them to uphold. This feels like a square peg in a round hole at the moment.

(Food for thought: maybe `Runtime` is a better place to handle concurrent agents in the same event loop? Or a new construct. We aren't limited to what we have now.)

academy/manager.py
```diff
@@ -55,14 +62,6 @@ class _RunSpec(Generic[AgentT, ExchangeTransportT]):
 async def _run_agent_on_worker_async(
```
benclifford: ^ this is now more like run agent in event loop, rather than "on worker", naming-wise

benclifford:
btw i tested this with my iterator prototype and it behaves as I expected.

gpauloski:
> the purpose of the Manager is to control the lifetime of Agents

It's semantics but rather than the purpose, I think this is a consequence of the fact the manager *owns* a set of executors that it runs agents on, and the executors have to be shut down when the manager's scope ends. 

I view the purpose of the manager as representing one or more execution domains within which agents can be run. (Agent lifetimes thus being associated with the lifetimes of the execution domains.) We blurred this by adding `event_loop` to the manager---the event loop execution domain has larger scope than the manager---but we sort of pretend that the manager owns an `event_loop` execution domain.

> I think launching "orphaned" agents should be extremely intentional

Totally agree. Part of why I brought that discussion up was that I think each scenario *does* need to be handled differently, and this solution mostly address 1 of the 4. That's not a bad thing but good to be thinking about how we will support the others and how our current APIs will need to evolve.

> why we should be concerned about multiple managers? 

The manager is a lightweight implementation---I'm not concerned about that---so it's more that it feels cognitively heavyweight. I think nested managers could make it harder for users to reason about their applications. (This is somewhat related to why I think circular imports and `TYPE_CHECKING` guards are often code smells.)

(1) Ownership of resources is blurred. Suppose I want a single executor to run all of my agents---even children. How do we control `num_agents <= num_workers` when there are multiple managers sharing resources? (Re: Ben blocking his GC endpoint and Yadu's deadlocks from the SC paper submission.)

(2) Error propagation is poorly defined. What should happen in the parent agent if it's child crashes?

With this implementation, the parent's manager will raise the exception caught in the child when the parent's manager is closed, which happens when the parent shuts down. That error raised during parent shutdown is then caught and raised when the user's manager shuts down. The errors get propogated at least, but I'd probably want a way for my parent agent to handle that error at runtime. The manager could handle this better, but at least when there's only one top-level manager you see all of the error logs and you get a nice exception group for all agents when the program exits.

The idiomatic way to do that in asyncio would probably be for `agent_run_alongside` to schedule and return a new `Task` for the child. The parent then keeps the reference to the child task and attached a callback to the `Task`. This is sort of what I was thinking when I mentioned maybe the `Runtime` should handle concurrent agents in the current event loop. Something like `self.agent_context.runtime.create_child(...) -> Task`.

(3) A consquence of (1) and (2) is that debugging becomes harder. Setting a breakpoint inside a top-level manager to inspect the running agents is now tricker because the top-level manager doesn't have a reference to all the agents in it's "domain".

To Ben's point about observing users for a bit, maybe we consider hiding the `agent_manager` property and keeping the use of a manager to create children in the same event loop as an internal implementation detail? It doesn't address all my concerns but at least we can have an interface for creating children without having to commit to the mechanisms. (Not that I mind breaking changes much at this stage.)

If we want to do that, it'd be good to rebase the PR. I have some comments but this includes the changes from #240. Agree with Ben on cleaning up `TYPE_CHECKING` guards as much as possible.

tox.ini
```diff
@@ -21,7 +21,7 @@ commands =
 extras = dev
 commands =
     coverage erase
-    coverage run -m pytest tests/integration {posargs}
+    coverage run -m pytest -s tests/integration {posargs}
```
gpauloski: Maybe revert?

academy/agent.py
```diff
@@ -716,6 +754,36 @@ def agent_shutdown(self) -> None:
...
         self.agent_context.shutdown_event.set()
 
+    async def agent_launch_alongside(
+        self,
+        agent: AgentT | type[AgentT],
+        *,
+        args: tuple[Any, ...] | None = None,
+        kwargs: dict[str, Any] | None = None,
+    ) -> Handle[AgentT]:
+        """Launch an child agent in the current event loop.
```
gpauloski: an -> a

academy/agent.py
```diff
@@ -716,6 +754,36 @@ def agent_shutdown(self) -> None:
...
+
+        Note:
+           This is a wrapper around `self.manager.launch`. For more options
+           and control over launching child agents, use `self.manager`
+           directly.
```
gpauloski: Remove this bit now to hide internal details

academy/agent.py
```diff
@@ -716,6 +754,36 @@ def agent_shutdown(self) -> None:
...
+
+        Args:
+            agent: Agent instance the agent will implement or the
```
gpauloski: I know this is copied from `Manager.launch` but it reads awkward. Maybe "An agent instance to launch, or an agent class to be instantiated before launching using `args` and `kwargs`."

tox.ini
gpauloski: I've removed all instances of tests printing to stdout so it doesn't really matter but sometimes those sneak in and clobber the CI test output of passing tests (stdout is still printed by pytest if a test fails)

gpauloski:
@AK2000 FYI I rebased this branch to fix the merge conflict I create from the proxystore PR

PR #248, opened 2025-10-31
Add an even more extra log mode targeted at observability

benclifford:
## Summary

extra becomes an int, a superclass of bool so existing True/False ints continue to work as before.

It's an open question if the user-facing documentation should switch to entirely conventional ints, rather than bools.

When a log file is configured, a second log file is created with a format of one JSON object per line. This is easy to work with using the `jq` command line tool and to read into Python for scripted processing.

The new JSON logs are fairly verbose: around 1kb per entry. This is a feature not a bug.

In the human-oriented logs, any logging extra keys beginning with "academy." are added onto the end of the human readable log lines.

This PR introduces a noqa: SIM115 which is advice that a file should be opened in a context manager. The logger arrangement isn't amenable to that. Further, logging is never closed explicitly and the log file is closed by end-of-process cleanup, so there isn't even an explicit close.

This PR adds a single example use of extras={"academy...."} annotation, at the start and end of an academy Manager. A separate PR will introduce more of that.

## Changes

## Testing
There's a new test case in the logging unit tests. That fails against current `main`, but PR #247 rearranges logging tests to provide more coverage.

## Pull Request Checklist

### DIFF
--- a/tests/unit/logging_test.py
+++ b/tests/unit/logging_test.py
@@ -14,7 +14,7 @@ from academy.logging import init_logging
 
 @pytest.mark.parametrize(('color', 'extra'), ((True, True), (False, False)))
 def test_logging_no_file(color: bool, extra: bool) -> None:
-    init_logging(color=color, extra=extra)
+    init_logging(color=color, extra=extra, force=True)
 
     logger = logging.getLogger()
     logger.info('Test logging')
@@ -30,7 +30,7 @@ def test_logging_with_file(
     tmp_path: pathlib.Path,
 ) -> None:
     filepath = tmp_path / 'log.txt'
-    init_logging(logfile=filepath, color=color, extra=extra)
+    init_logging(logfile=filepath, color=color, extra=extra, force=True)
 
     logger = logging.getLogger()
     logger.info('Test logging')


academy/logging.py
```diff
@@ -68,7 +93,7 @@ def init_logging(  # noqa: PLR0913
     logfile: str | pathlib.Path | None = None,
     logfile_level: int | str | None = None,
     color: bool = True,
-    extra: bool = False,
+    extra: int = False,
```
AK2000, line 96: Does it make sense to separate the two options (i.e. extra vs observability or some other naming)? To me it seems a little odd that extra includes both the information passed as a key word to log, and the thread/process information

AK2000:
Can you rebase this branch on `main` and remove the merge commit?

benclifford:
> Can you rebase this branch on main and remove the merge commit?

the github UI is only offering me "squash and merge" for merge right now which is going to lose this history anyway

academy/logging.py
benclifford: only weird if you have the prejudice that "extra" is secret code for "thread info", which most new users will not come with.

i don't really have a good feel for what the log config UI should look like: i don't want to recreate the entire Python logging configuration system which already exists and academy does not need to re-do badly, and I probably argue for *less* configurability in the helpers than *more* configurability.

AK2000: Yes, maybe the issue is calling thread info "extra", but I could imagine being frustrated if I pass "extra=True" and the "extra" dictionary that I put in my logging messages doesn't appear

benclifford: yeah there is a lot of scope for fiddling around there. I was trying to avoid all of that at this stage though so that Others can fiddle later.

gpauloski: `extra` could totally be renamed. I never really intended `init_logging` to be part of the public API so I put no thought into what the logging config should be. I just really needed a short cut for initializing logging across processes when I was debugging the original set of benchmarks/apps.

benclifford: ok I am not going to address this in this PR - we can all yak shave the config interface afterwards.

PR #252, opened 2025-11-04
Make cachetools required dependency

AK2000:
## Summary
Cachetools is required for server authentication, but is automatically imported no matter what. This optional dependency means that without `pip install academy[server]` things won't run. It does not seem like there is good reason to keep this as an optional dependency given that (1) it is alone, and (2) it is very light-weight, and (3) it is a well-maintained package that we have confidence in.

## Changes
- [x] Package (changes to package metadata or dependency versions)

## Testing

N/A

## Pull Request Checklist

- [ ] New and existing unit tests pass locally with the changes.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -33,6 +33,7 @@ dependencies = [
     "redis",
     "tomli ; python_version<'3.11'",
     "typing-extensions>=4.3.0 ; python_version<'3.11'",
+    "cachetools",
 ]
 
 [project.urls]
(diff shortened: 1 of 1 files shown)


benclifford:
orthogonal to this PR but on the same topic: I've had problems elsewhere with large accumulations of dependencies being automatically imported when importing something unrelated. Parsl's `import parsl` imports almost everything is one example of that.

Adding install dependencies fixes breakages on import but we should probably keep an eye on the longer term import load as cruft accumulates over time.

PR #254, opened 2025-11-04
Move proxystore to an optional dependency

gpauloski:
## Summary

Follow-up to the conversation in #252 about keeping import load minimal. ProxyStore is on the heavier side of the dependencies and only used in one module so I think it's safe to have as optional with a helpful import error message.

In addition, ProxyStore itself has lot's of extra install options so it's probably better to have users be a bit more deliberate about how they are installing ProxyStore. Academy only installs the base package which only supports 3 of the 10-ish connectors.

As a plus this might unblock Python 3.14 compatibility. Not sure when all of ProxyStore's dependencies will get updated.

I also sorted the dependency lists. Forgive me for including.

## Related Issues

- Related to #252

## Changes

- [x] Package (changes to package metadata or dependency versions)

## Testing

Tested in a virtual env w/o ProxyStore installed. You get this:

```bash
$ python
Python 3.12.10 (main, Apr  9 2025, 08:55:05) [GCC 11.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from academy.exchange import ProxyStoreExchangeFactory
>>> ProxyStoreExchangeFactory(None, None, None)
  File "/home/jgpaul/workspace/academy/academy/exchange/proxystore.py", line 14, in <module>
    from proxystore.proxy import Proxy
ModuleNotFoundError: No module named 'proxystore'

The above exception was the direct cause of the following exception:

  File "<stdin>", line 1, in <module>
  File "/home/jgpaul/workspace/academy/academy/exchange/proxystore.py", line 211, in __init__
    _assert_proxystore_available()
  File "/home/jgpaul/workspace/academy/academy/exchange/proxystore.py", line 45, in _assert_proxystore_available
    raise ImportError(message) from PROXYSTORE_IMPORT_ERROR
ImportError: Missing optional dependency proxystore. Install with:
  $ pip install academy-py[proxystore]  # as an extra
  $ pip install proxystore              # or proxystore[all]
```

## Pull Request Checklist

### DIFF
--- a/academy/exchange/proxystore.py
+++ b/academy/exchange/proxystore.py
@@ -38,11 +38,14 @@ except ImportError as e:  # pragma: no cover
 def _assert_proxystore_available() -> None:
     if PROXYSTORE_IMPORT_ERROR is not None:  # pragma: no cover
         message = textwrap.dedent(
-            """\
-            Missing optional dependency proxystore. Install with:
-              $ pip install academy-py[proxystore]  # as an extra
-              $ pip install proxystore              # or proxystore[all]""",
-        )
+            """
+            Missing optional dependency proxystore. Install using pip:
+              $ pip install proxystore
+            Or with extra options for specific connectors:
+              $ pip install proxystore[all]
+            Refer to the ProxyStore documentation for additional details.
+            """,
+        ).strip()
         raise ImportError(message) from PROXYSTORE_IMPORT_ERROR
 
 


AK2000 (dismissed):
This looks good to me.

`proxystore` is also used in the `pytest` suite, so if I forget to install `proxystore` and use `pytest` I don't see the informative error message: 
```
exchange/proxystore_test.py:9: in <module>
    from proxystore.connectors.local import LocalConnector
E   ModuleNotFoundError: No module named 'proxystore'
```
Does it make sense to put another informative message there, or should we just leave it because pytest is inherently internal (and you put proxystore in the tox.ini)?

gpauloski:
I'm indifferent on adding a message there.

Maybe it'd be better to skip the tests if proxystore isn't installed?

academy/exchange/proxystore.py
```diff
@@ -29,6 +24,27 @@
...
+except ImportError as e:  # pragma: no cover
+    PROXYSTORE_IMPORT_ERROR = e
+
+
+def _assert_proxystore_available() -> None:
+    if PROXYSTORE_IMPORT_ERROR is not None:  # pragma: no cover
+        message = textwrap.dedent(
+            """\
+            Missing optional dependency proxystore. Install with:
+              $ pip install academy-py[proxystore]  # as an extra
```
benclifford: elsewhere you indicated users might want to be a bit more thoughtful about which proxystore options they're installing - which maybe means `pip install proxystore[whatever]` is better advice (the 2nd option here) and academy packaging doesn't need to be involved at all.

gpauloski: That's reasonable to me. Do you think we should remove the `proxystore` optional dependency as well, or just change the error message? The alternative would be to add proxystore to the `dev` optional dependency group.

benclifford: sorry I didn't reply to this before.

i) does academy work without proxystore installed? (except for the proxystore tests)
ii) does academy work with proxystore, as tested by the proxystore tests?

That needs different install environments I think, but I don't have strong opinions on what those install environments look like. The test system is kinda setup for that already though.

In Parsl, that's a direction we've been slowly moving, driven by non-Python concerns because of our non-Python third party integrations with execution systems. Occasionally in the past we've failed to correctly test behaviour of non-optional dependencies because we only tested the "install-everything" environment.

gpauloski: Make sense. I think having an optional dependency group would make the testing of different environments easier. Putting `proxystore` in the `dev` won't give the granularity to test w/ and w/o.

AK2000: Does this mean that we should have separate CI tests for each optional dependency group?

benclifford: maybe. but a good start is: one env that is the bare-minimal academy, and another env that is everything-installed academy. later on, if there's one particular combination that keeps breaking for other people, then break that out into something more explicitly tested. there's 2^n combos if you want to go really overboard on the release engineering, but probably not right now.

gpauloski: I created a new issue to track this #279

PR #258, opened 2025-11-04
Test `JSONHandler` without `force=True` so pytest logger is not replaced

### LINKED ISSUE #257: Logging tests force replace pytest logger polluting remaining tests
### Describe the problem.

In #247 the logging tests were changed to set `force=True`. This replaces the pytest configured logging that is used for capturing the output when using `pytest --log-cli-level=INFO` so any tests that run after `tests/unit/logging_test.py` have no logging anymore. Instead, the logging just goes to stdouy, visible with `pytest -s`.

To reproduce, run the following and observe no logging outputs:
```
pytest --log-cli-level=INFO tests/unit/logging_test.py tests/unit/manager_test.py
```
If you remove `tests/unit/logging_test.py`, the logs come back.
```
pytest --log-cli-level=INFO tests/unit/manager_test.py
```

```console
From git
```

### Version

latest

### Python Version

3.12

### OS and Platform

x86 linux

gpauloski:
## Summary

#247 set the logging tests to use `force=True` so that the `JSONHandler` added in #248 would get properly covered. `force=True` replaced the pytest logger used for capturing, so the logging tests would inadvertently remove logging for all later test (described in more detail in #257).

Since the change to `force=True` was just about coverage, I added some unit tests for the `JSONHandler` and removed `force=True`. There was also an exception handle that was `#pragma: no cover` that I think was supposed to f-string format the exception but was missing the `f"` so I added a test for that.

## Related Issues

- Closes #257 

## Changes

*Not marking this as "bug" since it's just related to the tests and doesn't need to distract from the real bugs in the release notes.*

## Testing

Added new tests and confirmed logging works in pytest again.

## Pull Request Checklist

### DIFF
--- a/academy/logging.py
+++ b/academy/logging.py
@@ -183,8 +183,8 @@ class JSONHandler(logging.Handler):
         for k, v in record.__dict__.items():
             try:
                 d[k] = str(v)
-            except Exception:  # pragma: no cover
-                d[k] = 'Unrepresentable: {e!r}'
+            except Exception as e:
+                d[k] = f'Unrepresentable: {e!r}'
 
         json.dump(d, fp=self.f)
         print('', file=self.f)
--- a/tests/unit/logging_test.py
+++ b/tests/unit/logging_test.py
@@ -1,11 +1,13 @@
 from __future__ import annotations
 
+import json
 import logging
 import pathlib
 
 import pytest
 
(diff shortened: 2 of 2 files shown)


benclifford:
yes to the f-string change.

lol at trying to use unit test style for testing global environments

PR #266, opened 2025-11-07
Add groups support for the Cloud Exchange

yadudoc:
## Summary

This PR adds partial support for Agent sharing via Globus Groups. Here are the key changes:

* Added a dependent scope that allows Cloud Exchange to view groups and memberships of the clients
* Added `_get_dependent_tokens` that exchanges tokens for dependent tokens
* Added `_get_groups_and_memberships` that uses groups dependent tokens to query for membership info
* A new `group_info_cache_ttl_s` ttl which defaults to 5 minutes is added since querying groups is slower and group info changes infrequently. We could extend this even further.

Support for groups in the backend and user-facing methods to share an Agent will be added in follow-up PRs.

## Related Issues

N/A

## Changes

## Testing

[DRAFT] pending tests

## Pull Request Checklist

### DIFF
--- a/academy/exchange/cloud/authenticate.py
+++ b/academy/exchange/cloud/authenticate.py
@@ -97,7 +97,7 @@ class GlobusAuthenticator:
         *,
         token_cache_limit: int = 1024,
         token_ttl_s: int = 60,
-        group_info_cache_ttl_s: int = 60 * 5,
+        group_info_cache_ttl_s: int = 60,
     ) -> None:
         self._local_data = threading.local()
         self.executor = ThreadPoolExecutor(


academy/exchange/cloud/app.py
```diff
@@ -54,6 +54,7 @@
 from academy.exception import MessageTooLargeError
 from academy.exception import UnauthorizedError
 from academy.exchange.cloud.authenticate import Authenticator
+from academy.exchange.cloud.authenticate import ClientInfo
```
gpauloski: Should import directly from `client_info`

testing/redis.py
```diff
@@ -146,6 +148,17 @@ async def set(self, key: bytes | str, value: bytes | str) -> None:
...
 
+    async def sadd(self, key: bytes | str, *values: bytes | str) -> None:
+        key = self._encode(key)
+        if self.sets.get(key) is None:
+            self.sets[key] = set()
+
+        for value in values:
+            self.sets[key].add(self._encode(value))
+
+    async def smembers(self, key: bytes | str) -> builtins_set[bytes]:
```
gpauloski, line 159: Why does this need to be aliased rather than `-> set[bytes]`?

academy/exchange/cloud/authenticate.py
```diff
@@ -174,7 +197,76 @@ async def authenticate_user(self, headers: Mapping[str, str]) -> str:
...
+            ]
+        except KeyError:
+            logger.warning(
+                'Dependent tokens missing groups scopes %s',
+                dependent_tokens,
+            )
+            return group_info
+
+        response = requests.get(
+            url='https://groups.api.globus.org/v2/groups/my_groups',
```
gpauloski, line 255: Is there not a globus SDK interface for this?

academy/exchange/cloud/authenticate.py
```diff
@@ -174,7 +197,76 @@ async def authenticate_user(self, headers: Mapping[str, str]) -> str:
...
+                'Authorization': 'Bearer '
+                + str(groups_dep_token['access_token']),
+            },
+        )
+
+        if response.status_code != HTTPStatus.OK:
+            logger.error(
+                'Globus groups query failed: %d',
+                response.status_code,
+            )
```
gpauloski, line 266: style nit: prefer early returns (like above in `except`) and be explicit that the return is `return []` in the error case. This is easier to read and reduces the nesting required for the `else` statement which is the success case.

academy/exchange/cloud/backend.py
```diff
@@ -251,7 +311,60 @@ async def create_mailbox(
...
+        """Share a mailbox with a Globus group.
+
+        Args:
+            client: Client making the request.
+            uid: Target Mailbox for sharing
+            group_uid: Group id to share mailbox with.
+        """
+        if not self._has_mailbox_ownership(client, uid):
+            raise ForbiddenError(
+                'Mailbox sharing requires ownership',
```
gpauloski: This is vague. Maybe "Only mailbox owners can share a mailbox with a group."

academy/exchange/cloud/backend.py
```diff
@@ -251,7 +311,60 @@ async def create_mailbox(
...
+            )
+
+        if group_uid not in client.group_memberships:
+            raise ForbiddenError(
+                f'Owner does not belong to the group:{group_uid}',
```
gpauloski: If this error get's propagated back to the client, will it be clear who "Owner" is in this case?

Also nit: add space before group uid.

academy/exchange/cloud/backend.py
```diff
@@ -251,7 +311,60 @@ async def create_mailbox(
...
+            )
+
+        if uid not in self._shares:
+            self._shares[uid] = set()
+
+        self._shares[uid].add(group_uid)
+        logger.info('Mailbox:%s shared with group:%s', uid, group_uid)
```
gpauloski: Nit: replace the `:` with spaces. There's a few occurrences of this

academy/exchange/cloud/backend.py
```diff
@@ -81,7 +83,39 @@ async def create_mailbox(
...
+        uid: EntityId,
+    ) -> list[str]:
+        """Get list of globus groups the mailbox is shared with.
+
+        Only the owner of the Mailbox is allowed to share with a Globus Group.
+        This method should be idempotent.
+
+        Args:
+            client: Client making the request.
+            uid: Mailbox id to share.
```
gpauloski: The below implementation has a `Returns` section. Maybe include that here too.

academy/exchange/cloud/backend.py
```diff
@@ -251,7 +311,60 @@ async def create_mailbox(
...
+
+        Args:
+            client: Client making the request.
+            uid: Target Mailbox for sharing
+
+        Returns:
+            List of globus groups id strings
+        """
+        if uid not in self._owners:
+            raise BadEntityIdError(uid)
```
gpauloski: Should this be `ForbiddenError`?

gpauloski: Or is this being used as a proxy for if the mailbox exists?

academy/exchange/cloud/backend.py
gpauloski: I would suggest also documenting the errors that can be raised in all of these docstrings. I was a bit confused on this when reading the two implementations.

academy/exchange/cloud/backend.py
```diff
@@ -461,15 +574,90 @@ def _agent_key(self, uid: EntityId) -> str:
...
+
+        Args:
+             client: Client making the request.
+             uid: Mailbox id to get sharing info
+        """
+        owner = await self._client.get(
+            self._owner_key(uid),
+        )
+        if not owner:
+            raise BadEntityIdError(uid)
```
gpauloski: Same question as above about what's being checked here and what the error means.

academy/exchange/cloud/client.py
```diff
@@ -234,6 +234,102 @@ async def terminate(self, uid: EntityId) -> None:
...
+            ssl_verify = scheme == 'https'
+
+        session = aiohttp.ClientSession(
+            connector=aiohttp.TCPConnector(ssl=ssl_verify),
+            headers=connection_info.additional_headers,
+            trust_env=True,
+        )
+        return cls(session, connection_info)
+
+    def factory(self) -> HttpExchangeFactory:
```
gpauloski, line 282: If you use the factory to create the console, why would you need to use the console to create a new factory?

gpauloski (approved):
Not the most thorough review, but overall seems okay. My comments are mostly comments about something I didn't understand or style nits.

academy/exchange/cloud/backend.py
yadudoc: What do you think about “Mailbox sharing is restricted to mailbox owners.” ?

testing/redis.py
yadudoc: `mypy` gets confused because we have a method named set [here.](https://github.com/academy-agents/academy/pull/266/files/72651fc77f34267d703dc8dd9440862643f82393#diff-d6ff6655ffd2b63602c4450c105b3504ef5726fbf73bd94f2164e2498b9a3252R146)

academy/exchange/cloud/backend.py
yadudoc: Correct, it is a check for whether the mailbox exists. 

I just noticed that we don't do the same check for `share_mailbox`; can't share a mailbox that doesn't exist. It's cleaner if we broke the check for existence out of `_has_ownership` rather than assuming that we can claim ownership of any mailbox id that doesn't exist.

academy/exchange/cloud/backend.py
yadudoc: I added this one so that we can distinguish 1) mailbox does not exist vs 2) you don't have perms for the mailbox action.

academy/exchange/cloud/client.py
yadudoc: I'm hoping @AK2000 can chime in for this one.

AK2000: I think the bidirectionality is nice. For instance, if you pass the console around different places and then want to be able to create a manager or launch an agent from it. I've used the `ExchangeClient.factory` method similarly before. Of course you could extract the information to construct the factory manually, but I still think the convenience will be nice.

academy/exchange/cloud/backend.py
AK2000: I think in the other methods, we check if a mailbox exists by checking if it is in `self._mailboxes` and maybe we should continue this to be more consistent. 

For the redis backend it seems like I used the mailbox status. Maybe we switch the mailbox exists checks in the PythonBackend to be based on status, that way we get consistent behavior in terms of things like, what happens when you try to share a terminated mailbox?

academy/exchange/cloud/authenticate.py
AK2000: It looks like there is an SDK for this, but it makes the code much more verbose --- we have to create a new authorizer for each user and a groups client just to make this request. For now, I think I prefer using the API like this, but this would be easy to change if we find we need/want to do other groups queries in the future.

yadudoc: Correct. The SDK route is overkill. I've seen this pattern in several public gateway implementations that use Globus Auth.

PR #273, opened 2025-11-11
Adding guard to launching of background tasks

### LINKED ISSUE #244: Exception raised deep in response receipt causes hang without error report
### Describe the problem.

While investigating a serialization error, I came across a hang which I can reproduce with  this code change against `main` academy 12c69591b16642e34b75125b4071fb4e1e5b63b8.

This injected exception is representative of a real exception I've seen in serialization that I am investigating elsewhere and am not addressing in this issue.

I think this is a fairly standard "loop which must be running forever stops running forever, but nobody cares" bug that Parsl was/is full of.

```
diff --git a/academy/message.py b/academy/message.py
index f30fd94..fd078d7 100644
--- a/academy/message.py
+++ b/academy/message.py
@@ -409,6 +409,7 @@ class Message(BaseModel, Generic[BodyT]):
         Raises:
             TypeError: If the deserialized object is not a Message.
         """
+        raise RuntimeError("benc's injected exception")
         message = pickle.loads(data)
         if not isinstance(message, cls):
             raise TypeError(
```

then run tox and see that:

```
tests/unit/exchange/client_test.py ...............................FF.............^CROOT: [28943] KeyboardInterrupt - teardown started
ROOT: interrupt tox environment: py313
ROOT: requested interrupt of 29007 from 28943, activate in 0.00
ROOT: send signal SIGINT(2) to 29007 from 28943 with timeout 0.30
Task exception was never retrieved
future: <Task finished name='Task-588' coro=<ExchangeClient._listen_for_messages() done, defined at /home/benc/parsl/academy/academy/academy/exchange/client.py:197> exception=RuntimeError("benc's injected exception")>
  File "/home/benc/parsl/academy/academy/academy/exchange/client.py", line 200, in _listen_for_messages
    message = await self._transport.recv()
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/benc/parsl/academy/academy/academy/exchange/redis.py", line 196, in recv
    return Message.model_deserialize(raw[1])
           ~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^
  File "/home/benc/parsl/academy/academy/academy/message.py", line 412, in model_deserialize
    raise RuntimeError("benc's injected exception")
RuntimeError: benc's injected exception
^C

========================================================== FAILURES ===========================================================
_____________________________________ test_client_to_agent_message[HybridExchangeFactory] _____________________________________

```

This is reminiscent of a common cause of hangs in Parsl: some exception happens, falls off the top of the stack into not being reported.

```console
github, pytest
```

### Version

12c69591b16642e34b75125b4071fb4e1e5b63b8

### Python Version

Python 3.13.5 (main, Jun 25 2025, 18:55:22) [GCC 14.2.0] on linux

### OS and Platform

x86 linux

AK2000:
## Summary
When there are errors within Academy code (i.e. at places we don't expect to catch exceptions), exceptions can cause the entire academy program to hang without any meaningful errors or log messages. These errors are usually postponed until we `await` the task at the end of the program. 

This PR guards (almost) every launch of a background task with a wrapper that exits the program if an uncaught exception is raised. Taken largely from: https://github.com/proxystore/proxystore/blob/main/proxystore/utils/tasks.py. The goal is for bugs/breakages to be apparent immediately instead of things looking like they are working but potentially without one critical co-routine.

I was able to include this wrapper everywhere where a task was launched except for `Manager._run_agent` in `Manager.launch`. In this case, we don't have a clear separation between user code and Academy code. That is, an error in a user Agent looks the same at this point as an error within Academy. For instance, if a `Agent.agent_on_startup()` implementation raises an error, we want to retry or continue.  I think we can enforce a stricter distinction between errors raised in the `Runtime` and errors raised by the `Agent` which would make catching only `Agent` errors possible. However, I think this is an issue for another PR.
 
## Related Issues

- Closes #244

## Changes

## Testing

N/A

## Pull Request Checklist

### DIFF
--- a/academy/task.py
+++ b/academy/task.py
@@ -55,7 +55,7 @@ def spawn_guarded_background_task(
             are already logged by coro.
 
     Returns:
-        Asyncio task handle.
+        Asyncio task.
     """
     if log_exception:
         fut = asyncio.ensure_future(coro)


academy/event.py
```diff
@@ -21,7 +23,10 @@ async def wait_event_async(
         TimeoutError: If no event finished within `timeout` seconds.
     """
     tasks = {
-        asyncio.create_task(event.wait(), name=f'or-event-waiter-{i}'): event
+        spawn_guarded_background_task(
+            event.wait(),
+            name=f'or-event-waiter-{i}',
+        ): event
```
gpauloski, line 27: I don't think there is any value to adding this here. The `wait_event_async` is pretty narrowly defined and `Event.wait()` does not raise an exceptions.

academy/exception.py
```diff
@@ -155,6 +158,12 @@ def __reduce__(self) -> Any:
         return type(self), (self.aid,)
 
 
+class SafeTaskExitError(Exception):
```
gpauloski: I'd consider this an internal error meaning user code should never be importing this (either to raise it or to catch it) so this could be kept local to `academy.task`. It really only makes sense in the context of that module.

academy/logging.py
```diff
@@ -189,3 +194,18 @@ def emit(self, record: logging.LogRecord) -> None:
         json.dump(d, fp=self.f)
         print('', file=self.f)
         self.f.flush()
+
+
+async def execute_and_log_traceback(
```
gpauloski, line 198: Does this need to be public? Another case where if it's only used for `spawn_guarded_background_task` I think it should stay local to that module and be underscore-prefixed.

gpauloski: Oh I see you use it in the manager without `spawn_guarded_background_task`

academy/manager.py
```diff
@@ -494,7 +495,9 @@ async def launch(  # noqa: PLR0913
         )
 
         task = asyncio.create_task(
-            self._run_agent(executor_instance, spec),
+            execute_and_log_traceback(
```
gpauloski: I agree that we don't want to use `spawn_guarded_background_task` here but we already catch and log exceptions in `_run_agent` right?

academy/socket.py
```diff
@@ -344,7 +346,10 @@ async def _register_client_task(
         reader: asyncio.StreamReader,
         writer: asyncio.StreamWriter,
     ) -> None:
-        task = asyncio.create_task(self._handle_client(reader, writer))
+        task = spawn_guarded_background_task(
```
gpauloski, line 349: This might be another case where we end up with double-logged exceptions.

academy/task.py
```diff
@@ -0,0 +1,60 @@
...
+
+def spawn_guarded_background_task(
+    coro: Coroutine[Any, Any, Any],
+    *,
+    name: str,
+) -> asyncio.Task[Any]:
+    """Run a coroutine safely in the background.
+
+    Launches the coroutine as an asyncio task and sets the done
+    callback to [`exit_on_error()`][proxystore.utils.tasks.exit_on_error].
```
gpauloski: For simplicity, I'd suggest making renaming it `_exit_on_error` and just updating this docstring to say "sets the done callback to raise a [SystemExit][SystemExit] if the task raises an exception."

academy/task.py
```diff
@@ -0,0 +1,60 @@
...
+    This is "safe" because it will ensure exceptions inside the task get logged
+    and cause the program to exit. Otherwise, background tasks that are not
+    awaited may not have their exceptions raised such that programs hang with
+    no notice of the exception that caused the hang.
+
+    Tasks can raise [`SafeTaskExit`][proxystore.utils.tasks.SafeTaskExitError]
```
gpauloski: Update this to `academy.task.SafeTaskExitError`

tests/unit/task_test.py
```diff
@@ -0,0 +1,64 @@
...
+    async def run(task) -> None:
+        await spawn_guarded_background_task(
+            task(),
+            name='test-task',
+        )
+
+    with (
+        contextlib.redirect_stdout(
+            None,
+        ),
```
gpauloski: This is some funky indentation. I wonder why ruff doesn't collapse this to a single line

gpauloski (dismissed):
Maybe I was a little excited when I said we should replace all tasks with this :). There are a couple places where I don't think it's necessary and left some comments on locations for bits, but otherwise looks good.

academy/logging.py
```diff
@@ -189,3 +194,18 @@ def emit(self, record: logging.LogRecord) -> None:
...
+) -> None:
+    """Execute a coroutine and log any tracebacks.
+
+    Catches any exceptions raised by the coroutine, logs the traceback,
+    and re-raises the exception.
+    """
+    try:
+        return await coro
+    except Exception:
+        logger.error(traceback.format_exc())
```
benclifford: logger.exception("hello") will log the tracebook too, as will logger.error("hello", exc_info=True). both of those might be more structured approaches.

benclifford:
As an overall approach, deliberately trying to crash the program that is using your library is awful and doesn't scale well beyond toy examples that have no enclosing infrastructure.

On the flip side, raising SystemExit doesn't exit the program unless it is raised in very specific circumstances - only the main thread - which means this is going to work in simple cases and then as soon as someone does anything fun with threads, it will change behavior.

So maybe raise an AcademyIsBroken exception instead, which will exit the program anyway in almost the same circumstances, I think - the difference being that it will be more visible to users dealing with Exception.

academy/logging.py
AK2000: I'm going to leave this here because I do think it's a useful logging interface (i.e. if an agent wants to create a task).

gpauloski:
> As an overall approach, deliberately trying to crash the program that is using your library is awful

Totally agree. I think the idea here, to rephrase what @AK2000 said, is to fail as fast as possible if we already know there's a failure we can't recover from.

> SystemExit doesn't exit the program unless it is raised in very specific circumstances - only the main thread

Yeah, this approach wouldn't work with a concurrent.futures.Future b/c those callbacks run in separate threads, but callbacks on asyncio tasks/futures run in the main thread so it's okay in this case. 

> So maybe raise an AcademyIsBroken exception instead, which will exit the program anyway in almost the same circumstances, I think - the difference being that it will be more visible to users dealing with Exception.

I would be in favor of updating the logging in the callback to be more explicit about encountering a fatal error causing a system exit, but raising a `class AcademyIsBroken(Exception)` in the callback won't give the desired behavior of exiting the program at the point `AcademyIsBroken` is raised nor when the task is awaited (as @AK2000 mentioned).

Raising an `Exception` in a callback has non-intuitive behavior: the exception is caught and printed and then seemingly discarded (i.e., I don't think it's saved to the task object like `task.exception()` is). I played around with the following example but couldn't figure out how you could catch the `AcademyFatalError` from outside the callback, and it's not raised when awaiting the task---the exception from inside the task is what's raised.

```python
import asyncio

class AcademyFatalError(Exception):
    pass

def callback(task):
    if task.exception() is not None:
        raise AcademyFatalError()

async def task():
    raise RuntimeError('Oops!')

async def main():
    t = asyncio.create_task(task())
    t.add_done_callback(callback)
    print('task scheduled, now sleeping...')
    await asyncio.sleep(5)
    print('done sleeping')
    await t

if __name__ == "__main__":
    asyncio.run(main())
```

AK2000:
@benclifford I considered what you said, and added a "debug" option for exiting loops, so the SystemExit only happens when you start Academy (i.e. any Academy class) in "debug" mode.

The tests are failing because of an unrelated race condition in pinging agents that somehow was revealed by this PR.

AK2000:
Hmm, I still don't like the interaction with `asyncio`---what if someone is developing their own `asyncio` code and does not want the `academy` debug part? In terms of complexity, we still need to pass around debug at least in the `Manager` because we call `asyncio.run(...)` when launching a new agent in an executor.

What if we create a `academy.debug` variable, attached to an environment variable? That way you would get some of the simplicity back and still have separation.

academy/task.py
```diff
@@ -0,0 +1,70 @@
...
+
+    Args:
+        coro: Coroutine to run as task.
+        name: name of background task. Unlike asyncio.create_task, this is
+            required
+        log_exception: Write exception to the log. Set to false if exceptions
+            are already logged by coro.
+
+    Returns:
+        Asyncio task handle.
```
benclifford: this is an async `Task`, right? no handle involved?

PR #277, opened 2025-11-12
Tidy up informal schema around message-style (src/dest) logs

benclifford:
All message related extra logs now come from a helper attached to the Message class. That gives a logging API style that pushes responsibility onto an implementing class for describing what is interesting about itself (in the same way that `__repr__` does). I'm interested on review comments for that.

This makes all the source and dest fields now called `src` and `dest` without encoding some other behaviour-type specific name (like ping_dest).

Message-related logs now also include tag and label, to help correlating. This gives enough information to plot message flow diagrams from logs.

The source and destination metadata now come from the message object itself (because of the above API style). That is a change from before this PR, where some src/dest metadata was taken from other ambient variables. If the message metadata src/dest is the same as this other data, then that is fine, but I haven't reasoned much about if that is actually true.

## Changes

## Testing
visual inspection of log output

## Pull Request Checklist

- :fire: :boom:   Code changes pass `pre-commit` (e.g., ruff, mypy, etc.). :fire: no

### DIFF
--- a/academy/handle.py
+++ b/academy/handle.py
@@ -305,9 +305,8 @@ class Handle(Generic[AgentT]):
             exchange.client_id,
             self.agent_id,
             elapsed * 1000,
-            extra={
-                'academy.src': exchange.client_id,
-                'academy.dest': self.agent_id,
+            extra=request.log_extra()
+            | {
                 'academy.ping_time_s': elapsed,
             },
         )


academy/handle.py
```diff
@@ -310,8 +306,8 @@ async def ping(self, *, timeout: float | None = None) -> float:
             self.agent_id,
             elapsed * 1000,
             extra={
-                'academy.ping_src': exchange.client_id,
-                'academy.ping_dest': self.agent_id,
+                'academy.src': exchange.client_id,
```
AK2000: Why do we not use request.log_extra again here? It seems like the same situation as above?

benclifford: yeah that should work: "here's the response time for this ping message we sent before".

the code as I wrote it comes from me doing very localised "log stuff about the response" implementation, and there is no response `Message` here. but using the original `Message` as context makes sense.

PR #283, opened 2025-11-16
Check all imports can be imported at the top level

benclifford:
## Summary

A couple of times I've seen problems in the academy codebase to do with circular imports - most awfully, one passed CI and ended up in `main`, fixed by #281 

The implicit assumption that #281 tries to fix is that an academy module can be imported into an environment that has not been pre-disturbed by the earlier import of other academy stuff. That's a reasonable assumption.

Given the tangled nature of academy's mutual imports, I don't think #281 will be the last circular import problem, and while I think it is worth working towards uncircularising things, the code base is where it is.

As a comparison for what warty old codebases look like, I ran the test against the parsl repo, and encountered a handful of failures: mostly Python-based configuration files that aren't imported by the main codebase and make assumptions about the environment; but one slighly-suspicious import in the visualization code that assumes that some other web server context/environment has been established before the import is made. That one is ~50% legit in my eyes.

## Changes

- [X] Test (changes or additions to testing)

## Testing

This is a new test that passes in CI. I've tested that it really does fail, on my local system, with a suitably non-importable file, and seen it fail in a non-dev with mypy not installed.

## Pull Request Checklist

- [X] Relevant tags are added based on the types of changes.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -19,13 +19,13 @@ repos:
     hooks:
       - id: codespell
   - repo: 'https://github.com/astral-sh/ruff-pre-commit'
-    rev: v0.14.6
+    rev: v0.14.7
     hooks:
       - id: ruff-check
         args: ['--fix', '--exit-non-zero-on-fix', '--output-format', 'concise']
       - id: ruff-format
   - repo: 'https://github.com/pre-commit/mirrors-mypy'
-    rev: v1.18.2
+    rev: v1.19.0
     hooks:
       - id: mypy
         additional_dependencies:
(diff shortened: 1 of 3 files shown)


AK2000:
I like this a lot. Can/should we make this use the base academy install (instead of `dev, proxystore`)? This could help check that we do not unintentionally depend on optional packages especially when importing core classes (as was the case where #252 also pass CI). I'd like to do this, but it might mean that we need a "exclude" list of classes that require some optional dependency. Thoughts?

benclifford:
@AK2000 i should tidy it up a bit first, eg script in root of repo

AK2000 (approved):

PR #295, opened 2025-12-08
Add distributed log configuration system

benclifford:
## Summary

This PR introduces the concept of distributed Python `logging` configuration:

i) a user might want the same log configuration in several processes, because their workflow comprises several processes that should be observed together - an example would be a laptop-based workflow orchestration/submit side, and an HPC worker node based execution process.

ii) several authorities might regard themselves as the one entitled to configure logging for a Python process. An example here might be academy launching an agent in a remote Parsl worker inside a Globus Compute endpoint (and wanting to observe log information), and the enclosing Globus Compute endpoint also wanting to observe log information.

iii) users, implementers and researchers are interested in observing logging novel means: for example, under development distributed logging systems -- see diaspora or via systems focused on specific uses (for example) provenance - see flowcept.

These are handled quite badly/informally at the moment: Academy does some remote log configuration, but using logger.basicConfiguration which only happens if no one else has configured the root logger. That means it is very sensitive to what anyone else has done. The academy remote configuration is also restricted to a tiny language which can describe file targets on the remote file system and nothing else.

The notion of log configuration introduced in this PR is:

Log configurations are described by `LogConfig` objects, each of a class that knows how to configure logging in a particular way.

These configuration objects are expected to be serialized to other Python processes, primarily using pickle. In the presence of multiple serializations, these objects can still tell if they are from the same original source object (by use of an identifier field, rather than expecting the serialization system to preserve this identity relation).

These configuration objects are expected to know how to initialize their local process in their own particular way, on demand, and to know how to undo that configuration on demand.

None of these log configuration objects should expect to be the main log configuration for a process and should always expect to be alongside other configurations (either of this form, or set up by some other piece of Python code) - they are *visiting* a process in the sense that a Parsl task is a visitor inside a Parsl worker process , which has a separate existence (and so an academy worker inside a Parsl task is also a visitor).

Some academy scaffolding (`log_context`) manages these log configurations - for example as a context manager - to avoid duplicate initializations or too-early de-initializations.

This PR splits the log configuration from before this PR into a couple of different `LogConfig` implementations:

`ConsoleLogging` - which knows about logging to stdout, including with ANSI color.

`FileLogging` - which knows about logging to a file on the co-located filesystem.

There is a new `LogConfig` called `JSONPool` which knows how to log to an automatically named directory under the users home directory, in JSON format. This is intended for machine consumption.

There is a new `MultiLogConfig` which joins several `LogConfig` objects into one. This supports use cases such as logging to a file and also to some other experimental system simultanously.

Outside of this PR, I have also prototyped a `LogConfig` which knows how to send some logs into flowcept - https://github.com/ORNL/flowcept. I have also worked with Diaspora Octopus to drive some ideas for implementation, though I don't have a working implementation right now.

## Changes

## Testing

Added a bunch of tests around logging to the test suite.

## Pull Request Checklist

### DIFF
--- a/academy/logging/configs/jsonpool.py
+++ b/academy/logging/configs/jsonpool.py
@@ -1,0 +1,76 @@
+from __future__ import annotations
+
+import logging
+import pathlib
+import uuid
+from collections.abc import Callable
+
+from academy.logging.configs.base import LogConfig
+from academy.logging.helpers import JSONHandler
+
+logger = logging.getLogger(__name__)
+
+
+class JSONPoolLogging(LogConfig):
+    """Configures logging to files in home directory based pool of logs.
+
+    This feature is aimed at mechanical processing of logs, rather than
+    human readability.
+
+    Logs written by this configuration are stored in JSON format under
+    `~/local/share/academy/logs/`
+
+    Under that directory, logs are first separated into directories by
+    the (distributed) identity of the log configuration: logs configured
+    by the same JSONPoolLogging object, or by a serialized/deserialized
+    copy of the same object, will appear under the same directory.
+
+    Within that directory, each instance of log initialization will get a
+    new log file.
+
+    This logger is not configurable: it is intended to capture full debug
+    logs from the root, with selection of log records made during the
+    mechanical analysis stage.
+    """
+
+    def __init__(
+        self,
+    ) -> None:
+        super().__init__()
+
+    def init_logging(self) -> Callable[[], None]:
+        """Initialize JSON logging into shared pool."""
+        instance_id = str(uuid.uuid4())
+
+        # Home resolution is deferred until init_logging time because the path
+        # can be different on every invocation as the config object is moved
+        # between execution hosts.
+        path = (
+            pathlib.Path.home()
+            / 'local'
+            / 'share'
+            / 'academy'
+            / 'logs'
+            / self.uuid
+            / instance_id
+        )
+        path.parent.mkdir(parents=True, exist_ok=True)
+
+        json_handler = JSONHandler(path.with_suffix('.jsonlog'))
+        json_handler.setLevel(logging.DEBUG)
+
+        root_logger = logging.getLogger()
+        root_logger.addHandler(json_handler)
+        root_logger.level = min(root_logger.level, json_handler.level)
+
+        logger.info(
+            'Configured JSONPoolLogging (pool uuid=%s, path=%s)',
+            self.uuid,
+            path,
+        )
+
+        def uninitialize_callback() -> None:
+            root_logger.removeHandler(json_handler)
+            json_handler.close()
+
+        return uninitialize_callback
--- a/academy/logging/recommended.py
+++ b/academy/logging/recommended.py
@@ -1,0 +1,39 @@
+from __future__ import annotations
+
+import logging
+import pathlib
+
+from academy.logging.configs.base import LogConfig
+from academy.logging.configs.console import ConsoleLogging
+from academy.logging.configs.file import FileLogging
+from academy.logging.configs.multi import MultiLogging
+
+
+def recommended_logging(
+    level: int | str = logging.INFO,
+    *,
+    logfile: str | pathlib.Path | None = None,
+    logfile_level: int | str | None = None,
+    color: bool = True,
+    extra: int = False,
+) -> LogConfig:
+    """Returns a LogConfig that is hopefully useful to most users.
+
+    The config object can be passed to other environments to get the same
+    log configuration in those other environments.
+    """
+    configs: list[LogConfig] = []
+
+    # Always makes a console logger.
+    configs.append(ConsoleLogging(level=level, color=color, extra=extra))
+
+    # If a log file is specified, make human readable logger
+    if logfile is not None:
+        logfile_level = level if logfile_level is None else logfile_level
+        configs.append(
+            FileLogging(logfile=logfile, level=logfile_level, extra=extra),
+        )
+
+    lc = configs[0] if len(configs) == 1 else MultiLogging(configs)
+
+    return lc
--- a/academy/manager.py
+++ b/academy/manager.py
@@ -9,7 +9,6 @@ import warnings
 from collections.abc import Iterable
 from collections.abc import MutableMapping
 from concurrent.futures import Executor
-from concurrent.futures import ThreadPoolExecutor
 from types import TracebackType
 from typing import Any
 from typing import Generic
@@ -32,7 +31,8 @@ from academy.handle import exchange_context
 from academy.handle import Handle
 from academy.identifier import AgentId
 from academy.identifier import EntityId
-from academy.logging import init_logging
+from academy.logging.configs.base import LogConfig
+from academy.logging.helpers import log_context
 from academy.runtime import Runtime
 from academy.runtime import RuntimeConfig
 from academy.serialize import NoPickleMixin
@@ -54,9 +54,7 @@ class _RunSpec(Generic[AgentT, ExchangeTransportT]):
     agent_args: tuple[Any, ...]
     agent_kwargs: dict[str, Any]
     submit_kwargs: dict[str, Any]
-    init_logging: bool = False
-    loglevel: int | str = logging.INFO
-    logfile: str | None = None
+    log_config: LogConfig | None = None
 
 
 async def _run_agent_async(
@@ -89,16 +87,9 @@ def _run_agent_on_worker(
     academy_debug_mode: bool = False,
     **kwargs: Any,
 ) -> None:
-    if spec.init_logging:
-        if spec.logfile is not None:
-            logfile = spec.logfile.format(agent_id=spec.registration.agent_id)
-        else:
-            logfile = None
-
-        init_logging(level=spec.loglevel, logfile=logfile)
-
-    set_academy_debug(academy_debug_mode)
-    asyncio.run(_run_agent_async(spec))
+    with log_context(spec.log_config):
+        set_academy_debug(academy_debug_mode)
+        asyncio.run(_run_agent_async(spec))
 
 
 @dataclasses.dataclass
@@ -120,12 +111,6 @@ class Manager(Generic[ExchangeTransportT], NoPickleMixin):
         This class can be used as a context manager. Upon exiting the context,
         running agents will be shutdown, any agent handles created by the
         manager will be closed, and the executors will be shutdown.
-
-    Tip:
-        When using
-        [`ProcessPoolExecutors`][concurrent.futures.ProcessPoolExecutor],
-        use the `initializer` argument to configure logging in the worker
-        processes that will execute agents.
 
     Note:
         The manager takes ownership of the exchange client and executors,
@@ -159,6 +144,7 @@ class Manager(Generic[ExchangeTransportT], NoPickleMixin):
         *,
         default_executor: str = 'event_loop',
         max_retries: int = 0,
+        log_config: LogConfig | None = None,
     ) -> None:
         self._executors: dict[str, Executor | None] = {'event_loop': None}
         if isinstance(executors, Executor):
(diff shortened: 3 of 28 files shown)


academy/logging/configs/jsonpool.py
```diff
@@ -0,0 +1,122 @@
...
+
+    This is intended to make it so json logs go into a uniquely identified
+    filename, identified first by the instance of the file pool log, then
+    by a configuration invocation ID, so that we end up with a pool of logs
+    that are clearly related. Even if they are on different file systems,
+    in the distributed case, they should at least all appear on those file
+    systems under the same pool instance ID.
+
+    The logs are not configurable: they are full debug logs from the root,
+    in JSON format. That's intended to get *rid* of configurability for users
```
AK2000: I'm still not sure about this philosophy. We can by default set debug logs and then allow for a user to change the settings --- the idea being that a user who needs to configurability gets it, but otherwise we have full observability.

academy/manager.py
```diff
@@ -481,17 +477,25 @@ async def launch(  # noqa: PLR0913
...
+        #    executor_instance is None
+        #    or isinstance(executor_instance, ThreadPoolExecutor)
+        # ):
+        #    warnings.warn(
+        #        f'`init_logging` was specified for agent '
+        #        f'{registration.agent_id} running in the same process as the '
+        #        f'Manager. `init_logging` should only be called once per '
+        #        f'process.',
+        #        stacklevel=2,
+        #    )
```
AK2000: This was to avoid what I thought might be a common pitfall where people set `init_logging=true` when they didn't need to. `init_logging` multiple times per htex worker (or per process pool worker) is not much of an issue because they can override each other (i.e. the new agent running in that worker supplants anything else running in that worker, so the old thing doesn't log any more).

academy/logging/configs/jsonpool.py
```diff
@@ -0,0 +1,122 @@
...
+        # fancy stack inspection to # see who invoked init_logging to
+        # fabricate more human readable names, but a core json log notion is
+        # that you wouldn't be looking at individual files manually anyway.
+        instance_id = str(uuid.uuid4())
+
+        # home resolution is deferred until init_logging time because can
+        # be different on every invocation as the config object is moved
+        # between execution locations.
+        path = (
+            pathlib.Path.home()
```
AK2000: I could also imagine that hard coding the home directory could be a problem, for instance, on shared file systems where the home directory is by restricted to be tiny (this is the case on the CS cluster I sometimes use).

academy/exchange/cloud/app.py
```diff
@@ -486,9 +487,19 @@ def create_app(
 def _run(
     config: ExchangeServingConfig,
 ) -> None:
+    if config.log_config:
+        config.log_config.init_logging()
+
+    # TODO: in the parsl prototype, I pass in a bunch of context here
+    # and I wonder if that's sensible here too? parsl has at least the
+    # notion of a rundir, which academy does not, and also has named
```
AK2000: Good idea, I think something like a rundir would be useful here (i.e. to separate these logs from the other workflow logs).

AK2000 (commented):
This looks like a good direction to me. How do you picture this integrating with some distributed logging solution (that we talked about building upon rather than reimplementing). Were you thinking that those plug in as another `ObservabilityConfig`?

benclifford:
> This looks like a good direction to me. How do you picture this integrating with some distributed logging solution (that we talked about building upon rather than reimplementing). Were you thinking that those plug in as another `ObservabilityConfig`?

yeah -- @ryanchard and I have prototyped this a little bit as "just another Handler" that also needs some context/configuration about connecting to however the distributed logging system works -- eg server endpoint, credentials, whatever is relevant for that particular method.

academy/logging/configs/jsonpool.py
benclifford: When i was working on the tests, it seemed like it would be nice to focus the test log files into a test specific tmpdir, and I think that is broadly the same requirement. I'll add a base directory parameter. *but* the caveat is that you will need to be aware of everywhere that you are initializing logs: some random compute endpoint might not have a similar enough file system layout to what you imagine on the submit side. in which case: use default home, or implement your own custom configurator that knows how to figure things out for your specific circumstances.

(this was part of the mess around Parsl `Channel`s which made them awkward to configure and reason about)

academy/manager.py
benclifford: i think its broadly the right direction to go in - but you probably don't get to be the master log configurer for some python process that you are a guest in. so I'd like to think about what the semantics can be more seriously. For example, I don't want you screwing up GC worker logs just because you also want to have some agent logs: now you just destroyed any chance of getting user support from the GC team; nor do I want two different agents in the same event loop to be fighting over who owns process-wide logging.

academy/logging/configs/jsonpool.py
benclifford: i'm fine adding in configurability, it was mostly the lazy implementation option. mostly I was pushing towards the style of log everything and if you think you're such an advanced user you can turn this off, then we'll let you do that.

benclifford: I have added back a more conventional file logger which you configure like in-master init_logging logfile (with filename and log level explicitly specified)

That leaves the "pool of giant log files in home" logger separate, with the intention that people can experiment with it, as they might with any other log configuration/handler like e.g. octopus - separate from the more traditional log files.

I'll add in some configuration options for it too.

academy/exchange/cloud/app.py
benclifford: in a non-shared-fs distributed environment, the notion of a rundir is pretty poorly defined - or at least it was in Channels-era Parsl. So I'd like to not address anything like that in this PR, so I don't get sucked in.

academy/manager.py
benclifford: Mostly for my own reference as I compare against existing behaviour:

> init_logging multiple times per htex worker (or per process pool worker) is not much of an issue because they can override each other 

they don't override each other: the first thing which does a log init wins. which is sometimes not even academy. see how the log tests can't actually initialise logging because pytest got their first.

subsequent configs are ignored.

that's behaviour I'd like to change: if I ask for my agent logs to go to a file, I would like that to happen.

tests/integration/sequential_agent_logs_test.py
```diff
@@ -0,0 +1,111 @@
...
+            agent = LogAgent(a_str)
+            logger.info('prelaunch A')
+            handle = await manager.launch(
+                agent,
+                log_config=a_lc,
+            )
+            await handle.shutdown()
+            logger.info('after A shutdown')
+
+            # does this returning ^ mean that all shutdown activity (eg log
```
AK2000: No `handle.shutdown()` will wait until the `shutdown` request returns. You have to use `manager.wait()` to wait to ensure that an agent has actually shutdown.

academy/manager.py
benclifford: The (mis)behaviour that this warning is about should be gone now -- there can be multiple active/overlapping configs, and the tests in this PR test that.

academy/exchange/cloud/app.py
benclifford: I removed changes to cloud/app from this PR -- that application has enough of its own configuration system that I am leaving that to a separate PR.

tests/integration/sequential_agent_logs_test.py
benclifford: I used manager.wait (having forgotten about it twice...)

examples/01-actor-client/run-01.py
```diff
@@ -27,21 +27,21 @@ async def get_count(self) -> int:
 
 
 async def main() -> int:
-    init_logging(logging.INFO)
+    lc = recommended_logging()
+    with log_context(lc):
```
AK2000: I see this being an extremely common use case/pattern (i.e. a Manager nested in a log context, where the log config is then passed to every agent), but I don't really like the feel of having essentially the whole program nested within 2 context managers. What do you think about the manager accepting a log config, entering/exiting the log config as part of it's own context, and using the log config it was initialized with as a default to pass to all agents?

academy/logging/configs/jsonpool.py
```diff
@@ -0,0 +1,70 @@
...
+
+    def init_logging(self) -> Callable[[], None]:
+        """Initialize JSON logging into shared pool."""
+        instance_id = str(uuid.uuid4())
+
+        # Home resolution is deferred until init_logging time because the path
+        # can be different on every invocation as the config object is moved
+        # between execution hosts.
+        path = (
+            pathlib.Path.home() / '.academy' / 'logs' / self.uuid / instance_id
```
AK2000: I see creating this being difficult to parse --- the uuid of the entire log config is random and so is the instance id. When trying to analyze runs, how do I unpack which run and which instance I should look at? Could we create this folder (or maybe the entire log config?) with a date-time id (e.g. 20260514104200-uuid[:8]) or something like that to make distinguishing it easier? I guess my high level goal would be to easily associate directories with a specific run (and for a specific agent if we are creating different directories for each agent as I think would happen in this case).

AK2000: Can we also unify this path with the default token store path (which right now is ~/local/share/academy). I don't know which one should be preferred, but I don't like having two different meta-data directories for academy.

academy/logging/recommended.py
```diff
@@ -0,0 +1,39 @@
...
+from academy.logging.configs.multi import MultiLogging
+
+
+def recommended_logging(
+    level: int | str = logging.INFO,
+    *,
+    logfile: str | pathlib.Path | None = None,
+    logfile_level: int | str | None = None,
+    color: bool = True,
+    extra: int = False,
```
AK2000, line 18: I know you took this interface from the current `init_logging` but personally I don't want the `extra` logs to always come with the thread information (it doesn't hurt, but I haven't gotten a lot of use out of thread information). If we are re-doing this interface anyway, what are your thoughts on breaking these into different options?

AK2000 (dismissed):
Looks good to me. Minor comments (I think most if not all could be left to a separate PR if you want)

examples/01-actor-client/run-01.py
benclifford: yeah I think that tidies things up a lot, syntax-wise - I'll do that in this PR because it will also affect all the examples.

academy/logging/configs/jsonpool.py
benclifford: haha I was (at least when prototyping this) going the other way and *not* putting identities in the filenames, deliberately trying to model having eg a giant cloud hosted log pool that just stores a pile of log records and you have to figure this stuff out from the records themselves. which itself is probably influenced by not having a strong concept of a "run" when there's all different things with different lifetimes everywhere. and in that sense, you (a human) aren't meant to parse this - it's only to chop up the giant pool of records to avoid conflicting file names.

it should be pretty easy to timestamp/label that top directory so that a "run" is (e.g.) a creation of a log config (which is then passed to a manager and propagates, according to the other comment you made).

Naming the instance ID is something more complicated because log contexts don't have names as such -- but in the parsl version of this work, I have encountered something similar, so I think I'll have to figure this out. But I don't think it should block this #295 PR -- the jsonpool is in here as a demo to further develop rather than something I am treating as a finished feature.

benclifford: yep - I'll use the path that is already used in the existing `main` branch.

academy/logging/recommended.py
benclifford: I think its fine to break up the extra= parameter into specific feature flags.

examples/01-actor-client/run-01.py
benclifford: I've changed this. It looks much nicer!

academy/logging/recommended.py
benclifford: i'll do this in a subsequent PR

PR #300, opened 2025-12-12
Adding Action Cancellation

### LINKED ISSUE #299: Ability to Cancel Actions
### Describe the Request

Actions can be long running tasks, sometimes we find out the result is not needed before the action completes. The agent requesting an action should be able to cancel the action. This might require support both from the exchange and the runtime.

### Sample Code

```python

```

AK2000:
## Summary
This PR support action cancellation. 

Some design issues to note:
 * We assume a proper cancellation request comes after an action invocation. This imposes an ordering requirement on the exchange that is met by the current implementations but could be violated by certain distributed designs. However, this requires only client-centric consistency. An alternative would have been to save cancelled requests and check if a tag has been cancelled before executing, but I think this is unneeded complexity.
 * Cancellation responses are ignored by the handle. This has to do with the interface to the user --- the internals of the communication with the exchange are hidden by the handle object.
 * There were some odd race conditions around cancelling an action before it entered the `try...except asyncio.CancelledError`
 block, where a client would expect a reply but never gets one. Switching to eager task creation in python3.12 is the elegant solution but adding a `asyncio.sleep(0)` is a [documented solution](https://stackoverflow.com/questions/69145007/how-can-i-start-a-python-async-coroutine-eagerly).

## Related Issues

- Closes #299

## Changes

## Testing

N/A

## Pull Request Checklist

### DIFF
--- a/academy/manager.py
+++ b/academy/manager.py
@@ -356,7 +356,7 @@ class Manager(Generic[ExchangeTransportT], NoPickleMixin):
         run_count = 0
         retries = self._max_retries
 
-        while not self._closed.is_set():
+        while not self._closed.is_set():  # pragma: no branch
             run_count += 1
             if retries > 0:
                 retries -= 1


academy/exchange/client.py
```diff
@@ -205,6 +205,16 @@ async def terminate(self, uid: EntityId) -> None:
...
     async def _listen_for_messages(self) -> None:
+        # Transport recv does not necessarily wait on io and neither
+        # does _handle_message. If we are persistently receiving messages,
+        # this means the event loop might always be occupied by this task.
+        # In python >= 3.12, we use the eager task factory to avoid this.
+        # But before that, we need a sleep in this loop to yield the
+        # scheduler.
+        if sys.version_info >= (3, 12):  # pragma: >=3.12 cover
+            loop = asyncio.get_event_loop()
+            loop.set_task_factory(asyncio.eager_task_factory)
```
benclifford, line 216: this is a global state change for everything that uses tasks in the current thread's event loop, right? not constrained to the agent/exchange client that is doing the listening?

benclifford:
I would like to read more about the race condition you mentioned. Cancellation-style stuff has been super subtle race-wise in other projects I've worked on, and this should be heavily reviewed I think.

What was the race condition that you were experiencing? (aka. what two things were happening in the "wrong" order) and how did that race condition manifest to the user?

academy/exchange/client.py
AK2000: Yes. In python>=3.14, you can pass `eager_start` as an argument to `asyncio.create_task` and it would only affect the local state, but combining all versions (i.e. using `asyncio.sleep(0)`, `set_task_factory`, and `create_task`, looked pretty ugly)

AK2000:
> I don't get a feel (e.g. from the test cases) what the user-facing interface to cancellation is meant to look like.

For the user, this looks like just cancelling a Task, which they have always been able to do (see this [test case)](https://github.com/academy-agents/academy/blob/18f8b1602718471bd998fba9be6fc3329618eced/tests/unit/handle_test.py#L305). Except previously, this cancellation was silent to the Agent that was executing the action --- the action would be executed and the result returned and silently ignored. This PR allows the actual action to be cancelled, not just the handle invocation of the action.

> I would like to read more about the race condition you mentioned. Cancellation-style stuff has been super subtle race-wise in other projects I've worked on, and this should be heavily reviewed I think.
> 
> What was the race condition that you were experiencing? (aka. what two things were happening in the "wrong" order) and how did that race condition manifest to the user?

The race condition happened when a cancel request was sent immediately after an action request. Then, in the Runtime event loop, this [line](https://github.com/academy-agents/academy/blob/45a5d6165b218513335665cc9e5cca990c9e59d7/academy/runtime.py#L301) that cancels the action would happen before the start of [this](https://github.com/academy-agents/academy/blob/45a5d6165b218513335665cc9e5cca990c9e59d7/academy/runtime.py#L221) `try...except` block that catches action cancellation. Guaranteeing that the action starts means that we will not yield the scheduler till the first `await`, which is inside the block. This race condition would not manifest in a visible error to the user because the Task that was cancelled does not depend on receiving a reply. But from the perspective of the `Handle` there is a request to an `Agent` that was never replied to, but the `Agent` no longer has any record of that request because it has been cancelled. It would also break the test case that was explicitly listening for a reply.

PR #313, opened 2025-12-24
Fix Cancellation Race Condition in Handle

AK2000:
## Summary
As @benclifford found some race conditions in PR #300 after it was merged, this PR is to address those issues.
The added test case `test_client_handle_action_cancelled_during_send` reveals the race condition discussed in that PR where an action cancellation happens during  `await exchange_client.send` before entering the `try...except...` block in the `Handle.action` method. Moving the `send` into the `try...except` block fixes the problem.

From the discussion in that PR, if the cancellation happens during the send, we do not know weather the action message has actually been sent. We could use `asyncio.shield` to protect the send from cancellation, but upon reflection, I think this is unnecessary. If a cancellation happens before the `ActionRequest` is sent, the `Agent` receives a improper `CancellationRequest` which it replies to, and the `Handle` receives a `ActionInvalidStateError` which it ignores without affecting the user.

## Related Issues
N/A

## Changes

## Testing

N/A

## Pull Request Checklist

### DIFF
--- a/tests/unit/handle_test.py
+++ b/tests/unit/handle_test.py
@@ -1,10 +1,8 @@
 from __future__ import annotations
 
 import asyncio
-import contextlib
 import logging
 import pickle
-import sys
 from collections.abc import AsyncGenerator
 from typing import Any
 from unittest import mock
@@ -26,9 +24,11 @@ from academy.handle import ProxyHandle
 from academy.identifier import AgentId
 from academy.identifier import UserId
 from academy.manager import Manager
+from academy.message import ActionRequest
 from academy.message import CancelRequest
 from academy.message import ErrorResponse
 from academy.message import Message
+from academy.message import Request
 from academy.message import ShutdownRequest
 from testing.agents import CounterAgent
 from testing.agents import EmptyAgent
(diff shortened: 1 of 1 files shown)


tests/unit/handle_test.py
```diff
@@ -329,6 +331,44 @@ async def test_client_handle_action_cancelled_sends_request(
...
+
+    handle = await manager.launch(SleepAgent)
+    with mock.patch.object(
+        manager.exchange_client,
+        'send',
+        new_callable=mock.AsyncMock,
+    ) as send:
+
+        async def sleep(x: Any):
+            await asyncio.sleep(0.1)
```
benclifford: this kind of arbitrary time period sleep is usually a warning that some other race condition is being jiggled away without being fixed.

AK2000: This is only for the test suite. I needed a sleep here so that the task cancellation (line 361) is likely to happen during the `send` coroutine, which is what triggered the race condition in the previous code.

benclifford: look at PR #314 to be a bit more explicit about the sequencing of expected events.

PR #336, opened 2026-02-19
Remove untested log configuration

benclifford:
These tests don't need logging configured in a non-default way, so remove to make the tests slightly simpler.

This makes draft PR #295 a little simpler.

## Changes

## Testing

tox passes

## Pull Request Checklist

### DIFF
--- a/tests/unit/exchange/cloud/app_test.py
+++ b/tests/unit/exchange/cloud/app_test.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import logging
 import multiprocessing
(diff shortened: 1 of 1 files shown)


AK2000 (approved):

gpauloski:
I probably added that because pytest's logging capture doesn't work on the spawned child process, and I prefered it not to appear in the output `pytest -s`. So a pretty trivial reason, but if you see similar logging levels in other tests it's probably the same reason---keeping the pytest output less noisy.

Issue #33, opened 2025-02-06
Add notion of "inflight" messages


gpauloski:
### Describe the Request

Currently, an agent can pull a request from it's mailbox and then die during the request causing the request to get lost. Ideally, the request has some notion of "in-progress" and then "completed" where it is finally removed from the mailbox.

PR #376, opened 2026-04-07
Add notion of "inflight" requests #33

rajsuman31:
## Summary
Requests in the exchange are being tracked. Once a message reaches the exchange, we store it with the status "INFLIGHT". As soon as the exchange receives a response with the same message tag as that of the request, the request is marked as "COMPLETE".

## Related Issues

- Related to #33 

## Changes

## Testing

N/A

## Pull Request Checklist


academy/exchange/local.py
```diff
@@ -175,6 +182,42 @@ async def send(self, message: Message[Any]) -> None:
...
+                RequestStatus.INFLIGHT,
+                extra={
+                    'academy.message_tag': message.tag,
+                    'academy.src': message.src,
+                    'academy.dest': message.dest,
+                    'academy.status': RequestStatus.INFLIGHT,
+                },
+            )
+        elif message.is_response() and message.tag in self._state.requests:
+            self._state.requests[message.tag].status = RequestStatus.COMPLETED
```
AK2000: We probably want some sort of handling for when we get an unexpected response

academy/exchange/local.py
```diff
@@ -55,6 +57,16 @@ def __init__(self) -> None:
         self.queues: dict[EntityId, AsyncQueue[Message[Any]]] = {}
         self.locks: dict[EntityId, Lock] = {}
         self.agents: dict[AgentId[Any], type[Agent]] = {}
+        self.requests: dict[uuid.UUID, RequestInfo] = {}
+
+
+@dataclasses.dataclass
+class RequestInfo:
```
AK2000: This should probably be a common class if it is the same between multiple implementations of the exchange transport

academy/exchange/local.py
```diff
@@ -55,6 +57,16 @@ def __init__(self) -> None:
         self.queues: dict[EntityId, AsyncQueue[Message[Any]]] = {}
         self.locks: dict[EntityId, Lock] = {}
         self.agents: dict[AgentId[Any], type[Agent]] = {}
+        self.requests: dict[uuid.UUID, RequestInfo] = {}
```
AK2000: Can we use this dictionary in the `_respond_to_pending_requests_on_terminate` function so in-flight requests get a response?

academy/exchange/redis.py
```diff
@@ -86,6 +97,9 @@ def _agent_key(self, uid: AgentId[Any]) -> str:
     def _queue_key(self, uid: EntityId) -> str:
         return f'queue:{uid.uid}'
 
+    def _request_key(self, tag: uuid.UUID) -> str:
+        return f'request:{tag}'
```
AK2000: I think we'll need a way to look up pending requests by agent for when an agent is terminated. Maybe something like "request:{agent_id}:{tags}"?

academy/exchange/cloud/backend.py
```diff
@@ -661,6 +709,7 @@ def __init__(  # noqa: PLR0913
         )
         self.mailbox_expiration_s = mailbox_expiration_s
         self.gravestone_expiration_s = gravestone_expiration_s
+        self._requests: dict[uuid.UUID, RequestInfo] = {}
```
AK2000: For this one, we want the RedisBackend to be effectively stateless so that we can scale the number of workers

academy/exchange/redis.py
```diff
@@ -245,6 +256,49 @@ async def send(self, message: Message[Any]) -> None:
         elif status.decode() == _MailboxState.INACTIVE.value:
             raise MailboxTerminatedError(message.dest)
         else:
+            if message.is_request():
+                self._requests[message.tag] = RequestInfo(
```
AK2000: We should use redis to store this information instead of the dictionary

AK2000 (changes requested):
For testing, in `transport_test.py` can we add a case to `test_transport_terminate_reply_pending_requests` where we fetch a request from a mailbox and don't send a reply, and ensure that a `MailboxTerminatedError` is returned?

academy/request_state.py
```diff
@@ -0,0 +1,13 @@
+from __future__ import annotations
+
+import dataclasses
+
+from academy.identifier import EntityId
+
+
+@dataclasses.dataclass
```
AK2000: Do we need another class for this? Can we just use the header for this data structure?

academy/exchange/cloud/backend.py
```diff
@@ -474,6 +496,50 @@ async def put(self, client: ClientInfo, message: Message[Any]) -> None:
                 self.message_size_limit,
             )
 
+        if message.is_request():
+            self._requests[message.tag] = RequestInfo(
+                src=message.src,
+                dest=message.dest,
+            )
+            logger.info(
```
AK2000: Can we change these log messages to debug? I think it get's kind of verbose with the logs in `app.py` that say very much the same thing.

academy/exchange/cloud/backend.py
```diff
@@ -832,14 +907,44 @@ async def terminate(self, client: ClientInfo, uid: EntityId) -> None:
...
-                body = ErrorResponse(exception=error)
-                response = message.create_response(body)
-                with contextlib.suppress(Exception):
-                    await self.put(client, response)
+        messages: list[Message[Any]] = [
+            Message.model_deserialize(raw) for raw in pending
+        ]
+        requests: dict[uuid.UUID, RequestInfo] = {}
+
+        async for key in self._client.scan_iter(
```
AK2000: This seems like it might be a very costly scan? Can we offload more of this to redis by storing the requests keyed by destination?

academy/exchange/cloud/backend.py
```diff
@@ -832,14 +907,44 @@ async def terminate(self, client: ClientInfo, uid: EntityId) -> None:
...
+            'request:*',
+        ):  # pragma: no branch
+            tag_str = key.decode().split(':', 1)[-1]
+            info_data = await self._client.get(key)
+            info_dict = json.loads(info_data)
```
AK2000: Hmm is there a reason we need to use json here? I'd prefer to stick with `model_serialize` and `model_deserialize` for both consistency and validation.

academy/exchange/cloud/backend.py
```diff
@@ -968,6 +1073,62 @@ async def put(self, client: ClientInfo, message: Message[Any]) -> None:
                 self.message_size_limit,
             )
 
+        if message.is_request():
+            info_json = json.dumps(
```
AK2000: Same as json comment above. I think this would be easier using the `Header` to track requests instead of a custom class because the `header` is already a `BaseModel`

academy/exchange/transport.py
```diff
@@ -250,16 +255,52 @@ async def __aexit__(
 
 async def _respond_pending_requests_on_terminate(
     messages: Iterable[Message[Any]],
-    transport: ExchangeTransport[Any],
-) -> None:
+    send: Callable[[Message[Any]], Awaitable[None]],
+    requests: dict[uuid.UUID, RequestInfo] | None = None,
```
AK2000: I'm confused on the intended flow here? Isn't "requests" a superset of messages? How can any messages be in the mailbox that are not in requests.

tests/unit/exchange/cloud/backend_test.py
```diff
@@ -58,6 +59,7 @@ async def test_mailbox_backend_create_close(backend: MailboxBackend) -> None:
 
 @pytest.mark.asyncio
 async def test_mailbox_backend_send_recv(backend: MailboxBackend) -> None:
+    print('Testing send/recv...')
```
AK2000: Let's remove all of the print statements

tests/unit/exchange/cloud/backend_test.py
```diff
@@ -478,3 +480,204 @@ async def test_redis_backend_mailbox_expire(mock_redis) -> None:
...
+    await backend.put(client, message)
+
+    # Verify request is tracked as in-flight
+    assert message.tag in backend._requests
+    assert backend._requests[message.tag].src == uid
+    assert backend._requests[message.tag].dest == uid
+
+
+@pytest.mark.asyncio
+async def test_python_backend_request_tracking_completed() -> None:
```
AK2000: Is there a reason that the redis and python backed need separate tags?

tests/unit/exchange/transport_test.py
```diff
@@ -194,6 +196,60 @@ async def test_transport_terminate_reply_pending_requests(
                 await anext(listener)
 
 
+@pytest.mark.parametrize('factory_type', EXCHANGE_FACTORY_TYPES)
+async def test_pending_requests_on_terminate_returns_inflight_tags(
```
AK2000: I think this test should validate whether we get a response to a request event after the other agent has read that request from the mailbox. This looks to me like it's testing some internal methods rather than the external behavior

academy/exchange/local.py
```diff
@@ -57,6 +59,7 @@ def __init__(self) -> None:
...
```
AK2000: To clarify that comment about redis, it applies to local too. This dictionary could be {<destination_agent>: {<tag>: <request_header>} to avoid scanning through the dictionary when an agent is terminated

AK2000 (changes requested):
I think the functionality looks good, but I made a few style/structure comments

academy/exchange/cloud/backend.py
rajsuman31: Fixed.

tests/unit/exchange/cloud/backend_test.py
rajsuman31: done.

academy/exchange/cloud/backend.py
```diff
@@ -370,18 +369,19 @@ async def terminate(self, client: ClientInfo, uid: EntityId) -> None:
...
-        async with self._locks[uid]:
-            messages = await _drain_queue(mailbox)
-            for message in messages:
-                if message.is_request():
-                    error = MailboxTerminatedError(uid)
-                    body = ErrorResponse(exception=error)
-                    response = message.create_response(body)
-                    with contextlib.suppress(Exception):
-                        await self.put(client, response)
+        async def send(message: Message[Any]) -> None:
```
AK2000: Can we make this _send since this is a "private" method?

academy/exchange/cloud/backend.py
```diff
@@ -246,6 +244,7 @@ def __init__(
         self._terminated: set[EntityId] = set()
         self._agents: dict[AgentId[Any], tuple[str, ...]] = {}
         self._locks: dict[EntityId, asyncio.Lock] = {}
+        self._requests: dict[EntityId, list[Header]] = {}
```
AK2000: Can this be keyed on tag as well? So it becomes requests[destination][tag] = header?

academy/exchange/cloud/backend.py
```diff
@@ -665,6 +699,9 @@ def _queue_key(self, uid: EntityId) -> str:
     def _share_key(self, uid: EntityId) -> str:
         return f'share:{uid.uid}'
 
+    def _request_key(self, uid: EntityId) -> str:
+        return f'request:{uid.uid}'
```
AK2000: Same for this --- if we key on both uid and tag, we can easily find the messages that are pending both on terminate and on deleting messages.

academy/message.py
```diff
@@ -272,6 +272,27 @@ def create_response_header(self) -> Self:
             kind='response',
         )
 
+    @staticmethod
+    def list_serialize(headers: list[Header]) -> bytes:
```
AK2000: I don't think that we should need these methods. You should be able to use the default pydantic model_serialize and model_deserialize which will validate the fields when they are deserialized

tests/unit/exchange/transport_test.py
```diff
@@ -163,7 +163,7 @@ async def test_transport_terminate_reply_pending_requests(
     factory = get_factory(factory_type)
     async with await factory._create_transport() as transport1:
         async with await factory._create_transport() as transport2:
-            # Put a request and a response message in transport2 mailbox
+            # Send a request and a response message to transport2 mailbox
```
AK2000: This still does test if a response is sent once a request has been removed from a mailbox

AK2000 (changes requested):

tests/unit/exchange/transport_test.py
AK2000: I think if we write the tests in test_transport.py we can exercise all of the code paths without adding specific tests for each of the specific exchange transport implementations

academy/exchange/cloud/backend.py
rajsuman31: Technically, we can. But I think we may be going back and forth on assumptions. We started with tag as the key, then moved to dest, and now we want to use both as keys. Honestly, I feel the earlier design was better, as the cost to scan in Redis was one-time when the agent died, and we need requests filtered for that specific agent; unlike the current design, where we need to scan the requests every time a response is received and just discount the scan when the agent dies.

AK2000: But if it's destination and tag, we don't have to scan on response or when an agent dies

AK2000 (approved):

PR #379, opened 2026-04-07
Fix issue #366 - Heartbeat feature to keep track of last activity time

jz1909:
## Summary
Added a heartbeat feature that tracks the last activity time of mailboxes across all exchange transports

## Related Issues
#366 

## Changes

## Testing
N/A

## Pull Request Checklist

- [!] Code changes pass `pre-commit` (e.g., ruff, mypy, etc.).
- [!] New and existing unit tests pass locally with the changes.
- [!] Docs have been updated and reviewed if relevant.

### DIFF
--- a/academy/exchange/cloud/app.py
+++ b/academy/exchange/cloud/app.py
@@ -265,9 +265,7 @@ async def _terminate_route(request: Request) -> Response:
     return Response(status=StatusCode.OKAY.value)
 
 
-<<<<<<< HEAD
 @exception_to_response('discover')
-=======
 async def _get_heartbeat_route(request: Request) -> Response:
     data = await request.json()
     manager: MailboxBackend = request.app[MANAGER_KEY]
@@ -291,7 +289,6 @@ async def _get_heartbeat_route(request: Request) -> Response:
     return json_response({'heartbeat': heartbeat})
 
 
->>>>>>> f4f0916 (finished all transport layer heartbeat functionality)
 async def _discover_route(request: Request) -> Response:
     data = await request.json()
     manager: MailboxBackend = request.app[MANAGER_KEY]
@@ -389,12 +386,12 @@ async def _listen_mailbox_route(
     async with sse_response(request) as response:
         while response.is_connected():  # pragma: no branch
             try:
+                await manager.update_heartbeat(mailbox_id)
                 message = await manager.get(
                     client,
                     mailbox_id,
                     timeout=timeout,
                 )
-<<<<<<< HEAD
                 logger.info(
                     (
                         f'Fetched message {message.tag} from '
@@ -407,10 +404,6 @@ async def _listen_mailbox_route(
                         'academy.message_tag': message.tag,
                     },
                 )
-=======
-                logger.debug(f'Message at mailbox {message.dest} retrieved.')
-                await manager.update_heartbeat(mailbox_id)
->>>>>>> f4f0916 (finished all transport layer heartbeat functionality)
                 await response.send(message.model_dump_json())
             except (MailboxTerminatedError, TimeoutError) as e:
                 # These messages are not necessarily a sign something is wrong
@@ -446,6 +439,7 @@ async def _recv_message_route(request: Request) -> Response:
 
     client = get_client_info(request)
     try:
+        await manager.update_heartbeat(mailbox_id)
         message = await manager.get(client, mailbox_id, timeout=timeout)
     except MailboxTerminatedError:
         # We catch this exception separate from above and log it at a lower
--- a/academy/exchange/cloud/app.py
+++ b/academy/exchange/cloud/app.py
@@ -265,32 +265,15 @@ async def _terminate_route(request: Request) -> Response:
     return Response(status=StatusCode.OKAY.value)
 
 
+@exception_to_response('heartbeat')
 async def _get_heartbeat_route(request: Request) -> Response:
     data = await request.json()
     manager: MailboxBackend = request.app[MANAGER_KEY]
-
-    try:
-        raw_mailbox_id = data['mailbox']
-        mailbox_id: EntityId = TypeAdapter(EntityId).validate_json(
-            raw_mailbox_id,
-        )
-    except (KeyError, ValidationError):
-        logger.warning(
-            'Invalid or missing mailbox id in request.',
-            exc_info=True,
-        )
-        return Response(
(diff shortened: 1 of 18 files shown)


AK2000 (changes requested):
Some comments from our discussion:

- We ca n add `time` as a method to mock redis
- Every where we insert the heartbeat into redis we probably want to use redis.time to avoid any issues with clock synchronization

academy/exchange/cloud/app.py
```diff
@@ -515,6 +539,8 @@ async def _listen_mailbox_route(
                     mailbox_id,
                     timeout=timeout,
                 )
+                logger.debug(f'Message at mailbox {message.dest} retrieved.')
+                await manager.update_heartbeat(mailbox_id)
```
AK2000: I think this needs to be before the `get` otherwise the heartbeat is not updated unless there is a message to receive.

academy/exchange/cloud/app.py
```diff
@@ -599,17 +625,18 @@ async def _recv_message_route(request: Request) -> Response:  # noqa: PLR0911
...
-        extra={
-            'academy.message.event': 'GET',
-            'academy.message.src': message.src,
-            'academy.message.dest': message.dest,
-            'academy.message_tag': message.tag,
-        },
-    )
-    return json_response({'message': message.model_dump_json()})
+    else:
+        await manager.update_heartbeat(mailbox_id)
```
AK2000: Similarly, I think this needs to come before the `get`

academy/exchange/cloud/backend.py
```diff
@@ -228,6 +229,25 @@ async def remove_mailbox_shares(
         """
         ...
 
+    async def update_heartbeat(self, uid: EntityId) -> None:
+        """Update the heartbeat timestamp for a mailbox.
+
+        Args:
+            uid: Mailbox id to check.
```
AK2000: "Mailbox id to update."

academy/exchange/cloud/backend.py
```diff
@@ -522,6 +564,7 @@ async def share_mailbox(
             self._shares[uid] = set()
 
         self._shares[uid].add(group_uid)
+        logger.info('Mailbox %s shared with group %s', uid, group_uid)
```
AK2000: Maybe don't put these changes in this PR? I also think we do some logging at the app level, so this ends up being redundant.

academy/exchange/cloud/backend.py
```diff
@@ -381,6 +407,22 @@ async def terminate(self, client: ClientInfo, uid: EntityId) -> None:
                         await self.put(client, response)
 
             mailbox.shutdown(immediate=True)
+            logger.info(
```
AK2000: I think some of these comments were removed after the fork and may have gotten added back accidentally? Or maybe the branch hasn't been rebased?

academy/exchange/cloud/backend.py
```diff
@@ -841,6 +888,40 @@ async def terminate(self, client: ClientInfo, uid: EntityId) -> None:
                 with contextlib.suppress(Exception):
                     await self.put(client, response)
 
+    async def redis_current_time(self) -> float:
```
AK2000: Can we change this to `_redis_current_time` to indicate it is a helper method (and not supposed to be part of the public interface)?

academy/runtime.py
```diff
@@ -496,6 +498,11 @@ async def _start(self) -> None:
             name=f'exchange-listener-{self.agent_id}',
         )
 
+        self._heartbeat_task = spawn_guarded_background_task(
```
AK2000: Can we put this in the `ExchangeClient` like the listening loop? Then the heartbeat would be applicable to any mailbox not just agents/runtimes?

tests/unit/exchange/cloud/backend_test.py
```diff
@@ -478,3 +478,22 @@ async def test_redis_backend_mailbox_expire(mock_redis) -> None:
...
 
     await backend.terminate(client, uid)
+
+
+async def test_mailbox_backend_heartbeat(backend: MailboxBackend) -> None:
+    uid = UserId.new()
+    client = ClientInfo(str(uid), set())
+
+    heartbeat = await backend.heartbeat_status(uid)
+    assert heartbeat is None
```
AK2000: Should we have this raise a BadEntityIdError when the mailbox hasn't been created? That way it's differentiated from an agent that has been created but not been "started".

tests/unit/exchange/client_test.py
```diff
@@ -250,3 +252,30 @@ async def run():
...
+            dest=aid,
+            body=PingRequest(),
+        )
+        await client.send(message)
+    else:
+        await client._transport.update_heartbeat()
+
+    heartbeat = await client.heartbeat_status(client.client_id)
+    assert heartbeat is not None
+    assert heartbeat < 1.0
```
AK2000: For this assert, could you time the difference from when you call `send`/`update_hearbeat` to `heartbeat_status` and assert that the `heartbeat` is less than that time? That way we aren't sensitive to the heartbeat being fast or slow.

AK2000 (changes requested):

academy/exchange/cloud/backend.py
jz1909: Sorry, can you expand on this? Which comments are you specifically referring to?

jz1909: Wait, I don't think I added those - its probably what you said about rebasing

academy/exchange/cloud/backend.py
jz1909: I also didn't put this here, probably something wonky with rebasing as well

academy/exchange/cloud/app.py
```diff
@@ -265,6 +265,35 @@ async def _terminate_route(request: Request) -> Response:
...
+async def _get_heartbeat_route(request: Request) -> Response:
+    data = await request.json()
+    manager: MailboxBackend = request.app[MANAGER_KEY]
+
+    try:
+        raw_mailbox_id = data['mailbox']
+        mailbox_id: EntityId = TypeAdapter(EntityId).validate_json(
+            raw_mailbox_id,
+        )
+    except (KeyError, ValidationError):
```
AK2000: Minor: you can use the @exception_to_response decorator here instead of manually catching these exceptions

AK2000 (dismissed):
One minor comment, but looks good to me!

PR #382, opened 2026-04-08
Add Response on ShutdownRequest

AK2000:
## Summary
Add a response in the Runtime for shutdown requests. This simplified the handling of shutdown requests in the handle. To maintain the semantics of the current shutdown request, the acknowledgment is not waited for in `handle.shutdown`.

## Related Issues

- Closes #378 

## Changes

## Testing

N/A

## Pull Request Checklist

### DIFF
--- a/academy/handle.py
+++ b/academy/handle.py
@@ -109,7 +109,6 @@ class Handle(Generic[AgentT_co]):
             uuid.UUID,
             asyncio.Future[Any],
         ] = {}
-        self._shutdown_requests: set[uuid.UUID] = set()
 
         if self._exchange is not None:
             self._register_with_exchange(self._exchange)
@@ -167,33 +166,6 @@ class Handle(Generic[AgentT_co]):
         return remote_method_call
 
     async def _process_response(self, response: Message[ResponseT]) -> None:
-        # Check if this is an error response for a shutdown request as those
-        # are handled differently than other requests types (action, ping)
-        # which always expect a response.
-        if response.tag in self._shutdown_requests:
-            self._shutdown_requests.remove(response.tag)
-            body = response.get_body()
-            assert isinstance(body, ErrorResponse)
-            exception = body.get_exception()
-            # The only ok error to be ignored is if the agent we intended to
-            # shutdown was already shutdown.
-            if (
-                not isinstance(exception, AgentTerminatedError)
-                or exception.uid != self.agent_id
-            ):
-                logger.error(
-                    'Failure requesting shutdown for %s: %s (type: %s)',
-                    self.agent_id,
-                    exception,
-                    type(exception),
-                    extra={
-                        'academy.agent_id': self.agent_id,
-                        'academy.exception': exception,
-                        'academy.exception_type': type(exception),
-                    },
-                )
-            return
-
         future = self._pending_response_futures.pop(response.tag)
 
         if not future.cancelled():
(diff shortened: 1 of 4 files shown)


academy/handle.py
```diff
@@ -389,12 +361,33 @@ async def ping(self, *, timeout: float | None = None) -> float:
...
+    def _shutdown_callback(self, future: asyncio.Future[Any]) -> None:
+        if future.exception() is None:
+            return
+
+        exception = future.exception()
+        # The only ok error to be ignored is if the agent we intended to
+        # shutdown was already shutdown.
+        if (
+            not isinstance(exception, AgentTerminatedError)
+            or exception.uid != self.agent_id
```
benclifford, line 379: what's the circumstance that the uid wouldn't be the agent's uid here? (I see it's a move of code not new code so not a blocker for this PR)

academy/handle.py
```diff
@@ -389,12 +361,33 @@ async def ping(self, *, timeout: float | None = None) -> float:
...
+                    'academy.exception': exception,
+                    'academy.exception_type': type(exception),
+                },
+            )
+        return
+
     async def shutdown(self, *, terminate: bool | None = None) -> None:
         """Instruct the agent to shutdown.
 
-        This is non-blocking and will only send the request. Any unexpected
```
benclifford, line 401: this is still as true now as it was before?

academy/runtime.py
```diff
@@ -372,6 +372,10 @@ async def _request_handler(self, request: Message[Request]) -> None:
                 lambda _: self._action_tasks.pop(request.tag),
             )
         elif isinstance(body, ShutdownRequest):
+            response = request.create_response(SuccessResponse())
+            # We need to block here, because if we send this async,
+            # the exchange could be closed before the message is sent
```
benclifford, line 377: this comment is about when you start/end everything "workflow" style, right? with the same workflow submit script starting up an exchange and starting up agents. so this comment is talking about a race condition when that workflow submit script ends, right?

because in general the lifetime of agents and the lifetime of the exchange should be uncorrelated except by some other thing making them correlated.

if that's true, this brings up shutdown race condition vibes from parsl that i've spent years trying to untangle and which has contributed to a "of course parsl can't exit properly, don't expect it to" attitude amongst a whole load of people (myself included).

maybe thats an argument for making the shutdown call on the client side wait for the response, just like other calls - being even more consistent in behaviour than this PR pushes towards. This isn't doing any serious shutdown work before sending the response so shouldn't be more expensive than a regular action invocation?

(although along that lines, I see that the response goes back well before any shutdown even begins to happen, so there's separately another race condition if you're expecting your agent to do any nice shutdown behaviour like writing to disk before the workflow submit script perhaps hard terminates everything)

academy/handle.py
AK2000: Ah yes, did not mean to remove this comment. I debated between waiting for the future and not waiting for the future, and didn't get this comment right

academy/handle.py
AK2000: Not really sure, I guess if a message label gets messed up, so either a message gets delivered to the wrong handle or wrong agent?

academy/runtime.py
AK2000: > so this comment is talking about a race condition when that workflow submit script ends, right? because in general the lifetime of agents and the lifetime of the exchange should be uncorrelated except by some other thing making them correlated.

This comment is referring to a possible race condition between line runtime.py line 590 `await self._exchange_client.close()` which can be called after`signal_shutdown()` and the response being sent by the exchange client. 

>(although along that lines, I see that the response goes back well before any shutdown even begins to happen, so there's separately another race condition if you're expecting your agent to do any nice shutdown behaviour like writing to disk before the workflow submit script perhaps hard terminates everything)

There are two separate actions to be done: request shutdown, and wait for the agent to actually shutdown. The response just acknowledges that the shutdown request was received. You can separately wait for shutdown to finish, either by waiting for the future to return (like we do in the manager) or waiting till `run_until_complete` returns. This should handle the condition that your talking about since the `manager` will wait till all shutdown behavior finishes before terminating the agent, not just the shutdown request is acknowledged.

>maybe thats an argument for making the shutdown call on the client side wait for the response, just like other calls - being even more consistent in behaviour than this PR pushes towards. 

I thought about this a lot and might still be open to it (in a different PR), but for now I thought it would be good to preserve the semantics of the current method. I also don't know functionally if we should expect an agent to be online for a shutdown request to return without an error? Conceptually, unlike an action where we are expecting some result so it makes sense to wait until we get that result, I can see it being more common to just "tell" an agent to shutdown even if they are not online. Do we want to timeout or something like that if the agent is not online but the mailbox is still open --- that doesn't seem to make sense?

academy/handle.py
benclifford: ok. that's basically my understanding too (so I guess it could also be an assert because "something has gone wrong" if we hit that path)

academy/runtime.py
benclifford: ok let me poke about in the code a bit more to understand this

the use of exchange to also mean connection to/client of the exchange repeatedly confuses me (likewise with exchange factory which is a factory of not-exchanges)

PR #387, opened 2026-04-13
Batch Globus auth registration to reduce consent prompts.

GusEllerm:
## Summary

This change is local to the Globus exchange transport.

The function _create_registration was split between two phases / functions. First, _prepare_registration creates the globus client with its creds and scope requirements. This does *not* trigger user consent. Next, _finalize_registration authenticates all clients with a single user auth prompt.

This reduces the total number of consent prompts to 2 regardless of agent count. The first prompt is for the manage_projects scope, the second prompt is for all agent scopes batched into a single consent.

Changes:
- academy/exchange/cloud/globus.py: new _PendingRegistration dataclass,  _prepare_registration (phase 1: create client/cred/scope, accumulate scope requirements), _finalize_registration (phase 2: extract delegated tokens as cache hits), register_agents (orchestrates batch flow)
- academy/exchange/client.py: register_agents on UserExchangeClient with fallback to sequential registration for non-batch transports
- academy/manager.py: register_agents pass-through to exchange client
- examples/12-globus-exchange/run-12.py: example demonstrating batch registration with GlobusExchangeFactory

## Changes

## Testing

N/A

## Pull Request Checklist

### DIFF
--- a/academy/exchange/cloud/globus.py
+++ b/academy/exchange/cloud/globus.py
@@ -624,7 +624,7 @@ class GlobusExchangeTransport(ExchangeTransportMixin, NoPickleMixin):
 
         # Single login satisfies all accumulated scopes at once.
         assert self._app is not None
-        if self._app.login_required():
+        if self._app.login_required():  # pragma: no cover
             await loop.run_in_executor(
                 self.executor,
                 self._app.login,
--- a/academy/exchange/cloud/globus.py
+++ b/academy/exchange/cloud/globus.py
@@ -166,6 +166,21 @@ class AcademyGlobusClient(globus_sdk.BaseClient):
 
 
 @dataclasses.dataclass
+class _PendingRegistration(Generic[AgentT]):
+    """Intermediate state for a registration awaiting token delegation.
+
+    Created during the prepare phase of batch registration. Contains
+    the Globus Auth client, credential, and scope but not yet the
+    delegated token (which requires user consent).
+    """
+
+    agent_id: AgentId[AgentT]
+    client_id: uuid.UUID
+    secret: str
+    scope: Scope
+
+
+@dataclasses.dataclass
 class GlobusAgentRegistration(Generic[AgentT]):
     """Agent registration for hosted globus exchange."""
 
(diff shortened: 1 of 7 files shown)


academy/exchange/cloud/globus.py
```diff
@@ -540,6 +596,62 @@ async def register_agent(
...
+            List of agent registrations in order of input.
+        """
+        loop = asyncio.get_running_loop()
+
+        # first: create clients, credentials, scopes and
+        # add scope requirements without consent.
+        pending: list[tuple[_PendingRegistration[AgentT], type[AgentT]]] = []
+        for agent_type, name in agents:
+            aid: AgentId[AgentT] = AgentId.new(name=name)
+            p = await loop.run_in_executor(
```
AK2000: Minor thing, but can we gather all these futures and run them at once instead of doing this sequentially?

academy/exchange/cloud/globus.py
```diff
@@ -540,6 +596,62 @@ async def register_agent(
...
+                self.executor,
+                self._prepare_registration,
+                aid,
+            )
+            pending.append((p, agent_type))
+
+        # Single login satisfies all accumulated scopes at once.
+        assert self._app is not None
+        if self._app.login_required():  # pragma: no cover
```
AK2000: Instead of not covering this case, is there a way that we can write a test that mocks `self._app.login` and assert that it is only called once for this method? That way we can verify that we did the batching correctly (i.e. in case of regressions by potentially adding scopes or other things).

AK2000 (dismissed):
Looks good! Only a couple of minor comments

academy/exchange/cloud/globus.py
GusEllerm: `_prepare_registration` uses `self.auth_client`, which is a property that lazy inits a Globus `AuthClient`. On first access it checks `self.login_time`, sees its stale, and runs an interactive login. I think we discussed this -- the 30-min auth session. 

If we async gather these calls multiple threads enter the `auth_client` property at the same time. So each checks `self.login_time`, each sees stale, and each triggers its own interactive login. Then you get overlapping login requests in the terminal, clobbering each other. 

Also each `_prepare_registration` call mutates a shared `GlobusApp` state when adding scope requirements. I don't think (not 100%) that this is guaranteed thread-safe. E.g. a lost scope addition would case Phase 2 to silently fail to get a token. 

Ill have a think about how to handle this. Probably need to put a thread lock around the datetime check. Then I can refactor _prepare_registratoin to return the scope without applying it - that can happen after all scopes are ready.

GusEllerm: Done. When we create auth entities for each agent it is async:

```
prepared: list[_PendingRegistration[AgentT]] = list(
            await asyncio.gather(
                *(
                    loop.run_in_executor(
                        self.executor,
                        self._prepare_registration,
                        aid,
                    )
                    for aid in aids
                ),
            ),
        )
```

To make that safe a threading.Lock is added to the auth_client property. Without that each thread sees stale `login_time` and triggers the interactive prompt causing clobbering. 

Scope requirements are added external to `_prepare_registration` so that function does not modify the shared GlobusApp state from multiple threads. Scopes are added sequentially in `register_agents` after the gather returns.

The token and mailbox creation (phase 2) is also async now (this was easier).

academy/exchange/cloud/globus.py
GusEllerm: `test_register_agents_single_login` registers three agents and asserts that the batch login happens exactly once, and `add_scope_requirements` is called three times with the correct client id. 

This will need to change in the future when, I assume, we will add additional scopes for different capabilities to the agent harness. But it will fail loud in the meantime.

AK2000 (approved):

Issue #45, opened 2025-02-19
Launch agents as daemons


gpauloski:
Currently, launched agents are shutdown when the launcher is shutdown. Ideally, this would be an option so that a process can spawn an agent and exit without shutting the agent down.

PR #389, opened 2026-04-14
Launching Detached Agents

AK2000:
## Summary

Provides a mechanism to launch agents that automatically detach from a manager. I anticipate this is going to generate some discussion. I decided to use subprocess.Popen here to create a child process, and used `start_new_session` to free it from its parent process --- not sure if more should be done (i.e. to set the group or make process a daemon). This process reads the spec from standard in to load the agent for the first time. The retry and error catching mechanism of the manager are applied to the instantiation of the child process, but are not effective after that. This is dangerous (but necessary?) --- bugs and errors raised by the child are not propagated anywhere. Serialization becomes an immediately became an issue here, the `testing` package used in pytest causes errors when used in the child process. As an idea, I moved the `testing.agents` module into the `academy` package --- we could try to find another way around this (i.e. "mainify" an agent class for testing) or we could lean in fully and move the entire testing package to part of the `academy` package.

Other ideas worth mentioning/discussing:
- For the specific case of Globus Compute (which I anticipate is a common use case for this pattern) futures can outlive the executor naturally. In that case, all we really need to do is delete/ignore the future from the `manager` in order to detach the agent. I went with th a general solution that could be used for all executors but we could tailor this more to globus compute.
- The agent registration is not saved anywhere for restarting the agent, nor is the process id saved creating a mysterious process that is not kept track of which feels wrong --- we could potentially save these to a `local` folder enabling a restart mechanism (combined with the heartbeat PR #379)
- The add `remote.py` script ideally can be reused without the manager to launch agents as daemons/systemd units (#45)

This is part of a larger push toward long-lived agents which should be kept in mind when thinking about the ergonomics of this functionality.

## Related Issues

See discussion above.

## Changes

## Testing

N/A

## Pull Request Checklist

### DIFF
--- a/academy/manager.py
+++ b/academy/manager.py
@@ -4,6 +4,8 @@ import asyncio
 import contextlib
 import dataclasses
 import logging
+import pickle
+import subprocess
 import sys
 import warnings
 from collections.abc import Iterable
@@ -57,6 +59,7 @@ class _RunSpec(Generic[AgentT, ExchangeTransportT]):
     init_logging: bool = False
     loglevel: int | str = logging.INFO
     logfile: str | None = None
+    detached: bool = False
 
 
 async def _run_agent_async(
@@ -101,12 +104,36 @@ def _run_agent_on_worker(
     asyncio.run(_run_agent_async(spec))
 
 
+def _run_agent_detached(
+    spec: _RunSpec[AgentT, ExchangeTransportT],
+    academy_debug_mode: bool = False,
+    **kwargs: Any,
+) -> int:
+    args = [
+        sys.executable,
+        '-m',
+        'academy.remote',
+    ]
+    if academy_debug_mode:
+        args.append('--debug')
+    proc = subprocess.Popen(
+        args,
+        stdin=subprocess.PIPE,
+        start_new_session=True,
+    )
+    assert proc.stdin is not None  # Necessary for mypy
+    pickle.dump(spec, proc.stdin)
+    proc.stdin.close()
+    return proc.pid
+
+
 @dataclasses.dataclass
 class _ACB(Generic[AgentT]):
     # Agent Control Block
     agent_id: AgentId[AgentT]
     executor: str
     task: asyncio.Task[None]
+    detached: bool = False
 
 
 class Manager(Generic[ExchangeTransportT], NoPickleMixin):
@@ -276,7 +303,7 @@ class Manager(Generic[ExchangeTransportT], NoPickleMixin):
         self._closed.set()  # Stop from launching retries
 
         for acb in self._acbs.values():
-            if not acb.task.done():
+            if not acb.task.done() and not acb.detached:
                 handle = self.get_handle(acb.agent_id)
                 with contextlib.suppress(AgentTerminatedError):
                     await handle.shutdown()
@@ -345,7 +372,7 @@ class Manager(Generic[ExchangeTransportT], NoPickleMixin):
         self._default_executor = name
         return self
 
-    async def _run_agent(
+    async def _run_agent(  # noqa: C901
         self,
         executor: Executor | None,
         spec: _RunSpec[AgentT, ExchangeTransportT],
@@ -385,7 +412,22 @@ class Manager(Generic[ExchangeTransportT], NoPickleMixin):
             )
 
             try:
-                if executor is not None:
+                if spec.detached and executor is not None:
+                    await asyncio.wrap_future(
+                        executor.submit(
+                            _run_agent_detached,  # type: ignore[arg-type]
+                            spec,
+                            **spec.submit_kwargs,
+                        ),
+                        loop=loop,
+                    )
+                elif spec.detached and executor is None:
+                    await loop.run_in_executor(
+                        None,
+                        _run_agent_detached,
+                        spec,
+                    )
+                elif executor is not None:
                     await asyncio.wrap_future(
                         executor.submit(
                             _run_agent_on_worker,  # type: ignore[arg-type]
@@ -404,23 +446,18 @@ class Manager(Generic[ExchangeTransportT], NoPickleMixin):
                 )
                 raise
             except Exception:
+                logger.exception(
+                    'Received exception from %s, retries remaining: %d',
+                    agent_id,
+                    retries,
+                    extra={'academy.agent_id': agent_id},
+                )
                 if retries == 0:
-                    logger.exception(
-                        'Received exception from %s',
-                        agent_id,
-                        extra={'academy.agent_id': agent_id},
-                    )
                     if spec.config.terminate_on_error:
                         await self.exchange_client.terminate(
                             spec.registration.agent_id,
                         )
                     raise
-                else:
-                    logger.exception(
-                        'Restarting %s due to exception',
-                        agent_id,
-                        extra={'academy.agent_id': agent_id},
-                    )
             else:
                 logger.debug(
                     'Completed %s task',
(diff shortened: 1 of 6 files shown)


benclifford:
not looking at the code, but addressing your PR description:

> The add remote.py script ideally can be reused without the manager to launch agents as daemons/systemd units (https://github.com/academy-agents/academy/issues/45)

My gut says "system administrators know how to run daemons already" and that this could be the main focus of this PR - we shouldn't be trying to reinvent "lame systemd".

Most of my comments below are made with that worldview - basically ignoring any idea that there's a submitting workflow-style process that launches agents that then detach

(in part because that feels to me like its jumbling up the lifetime of the workflow-style bit vs the lifetime of the agents bit).

> The retry and error catching mechanism of the manager are applied to the instantiation of the child process, but are not effective after that. This is dangerous (but necessary?) --- bugs and errors raised by the child are not propagated anywhere

In a "sysadmins know how to run daemons" worldview: with systemd you do things like "log to stderr then exit" and let systemd handle status reporting just like any other daemon - or whatever daemon management system you use. (even "run a `/usr/bin/screen` session and run the process in there, which is a legit low end way of running things).

With that orientation, the important thing is to look like everyone else rather than to write mechanisms to deal with processes - anything to do with daemonisation/process groups/etc is the business of systemd or whatever the user chose to use, not academy.

In the observability worldview, you "log to wherever the user wanted to log" - aka #295 - which aligns with "log to stderr and exit".

> Serialization becomes an immediately became an issue here, the testing package used in pytest causes errors when used in the child process. As an idea, I moved the testing.agents module into the academy package --- we could try to find another way around this (i.e. "mainify" an agent class for testing) or we could lean in fully and move the entire testing package to part of the academy package.

how does this mirror what a user would see with their own applications? // how is this catching a problem that a user will see when they try to use academy?

if you regard the testing code as "user application" code, the user can't move their application code into the `academy.*` package. or at least, we prefer them to not do so.

When I've done stuff with my own toy agents suite, I've had to structure my code in certain ways, with some classes defined in importable/imported packages, etc, to make serialization work. If that's our solution/advice for a user experiencing serialization problems(?), then we should take that advice here as much as possible.

> The agent registration is not saved anywhere for restarting the agent, nor is the process id saved creating a mysterious process that is not kept track of which feels wrong

In a systemd first worldview: the "mysterious process" is tracked by systemd. This ties in to my comment above about jumbling up the lifetime of the workflow-ish bit vs the lifetime of the agent.

In any way of launching, if we want agent shutdown->restartability, saving enough info to restart sounds like something to implement soon.

Trying to use task-oriented Globus Compute to do "permanently alive" services sounds like a not-GC usecase (that GC people might want to support, but that's not how its architected). As an off the cuff idea, having a systemd-style permanent process somewhere like a login node that then sits and watches GC to make sure it's alive seems like a weird way to do it. Might be interesting to involve GC people in this?

AK2000:
> My gut says "system administrators know how to run daemons already" and that this could be the main focus of this PR - we shouldn't be trying to reinvent "lame systemd"

My thought behind remote.py was that it would become the target of the systemd service (i.e. users would create an agent registration and save it to their directory, and instead of having to right their own script to start the agent, they just use remote.py --- and we support this pathway with documentation/instructions). I think it's relatively easy for people to write their own script that starts an agent (which is why remote.py is less than 50 lines long), but it could still provide convenience and reduce boiler plate.

However, I  still think there is value to be had from being able to (remotely) launch agents that live past the lifetime of the script that starts them. The target here is "long-lived fire and forget agents". `systemd` is still a fairly heavy/technical lift to set up when compared to the target ease of use of tools like globus compute and Academy (with the essential premise being that maintaining ssh tunnels is hard). 

Maybe the right approach here is the `_launch_agent_detached` function builds and starts a systemd service? That would 1) give us the easy fire and forget functionality that I'm looking for, 2) avoid rebuilding a lame systemd and instead use something that works and 3) allow us to manage logs and restarts well. However I don't know if this is great practice to just be creating systemd services willy-nilly and I don't know if it is general enough (there were concerns with some systems not supporting user-space systemd processes)?

AK2000:
> When I've done stuff with my own toy agents suite, I've had to structure my code in certain ways, with some classes defined in importable/imported packages, etc, to make serialization work. If that's our solution/advice for a user experiencing serialization problems(?), then we should take that advice here as much as possible.

I think this is what we want: agents must be in importable packages to make serialization work. I guess the root of my question here is because I don't understand how the packaging is working right now. Why, when I run `pytest` can the code `import testing`, but if I try something like `uv run python conftest.py` I get an import error for the testing package. My thought here is that things in the testing package should be serializable. One way to accomplish this would be to move the testing package inside the academy folder (similar to the structure of parsl and globus_sdk where the testing code is part of the package), but is there a more elegant way by adjusting `pyproject.toml`?

benclifford:
> My thought behind remote.py was that it would become the target of the systemd service (i.e. users would create an agent registration and save it to their directory, and instead of having to right their own script to start the agent, they just use remote.py --- and we support this pathway with documentation/instructions). I think it's relatively easy for people to write their own script that starts an agent (which is why remote.py is less than 50 lines long), but it could still provide convenience and reduce boiler plate.

right - I mean to give the impression I was agreeing with this piece.

benclifford:
> However, I still think there is value to be had from being able to (remotely) launch agents that live past the lifetime of the script that starts them. The target here is "long-lived fire and forget agents".

The "forget" part of the "fire and forget" buzz usually turns out to be "fire and someone else cares about this", not "forget" so I feel that wording is cognitively dangerous. We need to be choosing who cares about the agent, not forgetting about this.

Lifetime-wise I think the "forget" model pretty immediately runs into things like:

i) my agent ended for one of various expected reasons, how do I make it restart?
ii) my agent deliberately shuts itself downwhen its idle, how do I make it restart "on load" - eg when someone sends it a message or when someone starts a new workflow and it isn't already running.
iii) i want to know if my agents are running

and that leads to having a supervisor of some kind to supervise those agents, and that supervisor probably definitely *doesn't* want to forget the agent it launched. and that supervisor is not magically going to exist for just the right lifetime the user wants.

benclifford:
> Maybe the right approach here is the _launch_agent_detached function builds and starts a systemd service? That would 1) give us the easy fire and forget functionality that I'm looking for, 2) avoid rebuilding a lame systemd and instead use something that works and 3) allow us to manage logs and restarts well. However I don't know if this is great practice to just be creating systemd services willy-nilly and I don't know if it is general enough (there were concerns with some systems not supporting user-space systemd processes)?

This is going to be different for each use case I think - we can (and should) template up userspace systemd for users that want that to exist, but in the end they are going to have to (if they're using it) acknowledge that systemd exists.

But there are plenty of other use cases that I think raw agent-in-systemd isn't going to support - e.g. keep alive in GC forever; and the load based use cases that I mentioned in another comment and that I think @wardlt mentioned in a talk the other day. None of these need detachability?

benclifford:
do you have a CI test you can commit that is failing because of serialization?

gpauloski:
Generally agree with the discussion so far. (Not looked at code yet either; happy to when you are ready for it.)

I see value in providing an entry point to running an agent without all the fluff and that can be easily called by other systems/tooling. Though, "remote" seems an odd name to me. I'd picture something like `python -m academy.run /path/to/my/agent/spec`. (With the spec ideally not being pickled objects, something more like a proper config filetype. I've wondered about this potential need before.)

> and that leads to having a supervisor of some kind to supervise those agents

This came up often last summer (and is mentioned in the linked issue): the desire for some service that will manage the lifetimes of your agents for you across some set of endpoints (Globus Compute, etc.) that you own. That could also justify some sort of actor support within Globus Compute.

AK2000:
Okay, I think my plan will be to break out the `remote` script in to a different PR with accompanying documentation since this is a component we can all agree on.

I still think we need an easier mechanism to launch these persistent agents than just the script. Since the primary use case for this is with globus compute, if we just delete the future from the `manager` this might be sufficient for the majority of use cases.

PR #390, opened 2026-04-14
Default Timeout for Cloud Exchange Listen and Receive

### LINKED ISSUE #386: Out of Resources in Http Exchange App, File Leaks
### Describe the problem.

When running the HttpExchangeServer for a long period of time, I see this error flood the logs:

```
[2026-04-12 10:25:54.428] ERROR    (asyncio) socket.accept() out of system resource
socket: <asyncio.TransportSocket fd=5, family=2, type=1, proto=0, laddr=('127.0.0.1', 8700)>
  File "/usr/lib/python3.12/asyncio/selector_events.py", line 178, in _accept_connection
    conn, addr = sock.accept()
                 ^^^^^^^^^^^^^
  File "/usr/lib/python3.12/socket.py", line 295, in accept
    fd, addr = self._accept()
               ^^^^^^^^^^^^^^
OSError: [Errno 24] Too many open files
```

This error seems to appear and recover transiently. However, when it appears, it can flood the logs --- I see a log message every ~0.2 ms, causing a tremendous amount of disk space use.

```console
pip install <from git>
gunicorn to run
```

### Version

v0.4.0

### Python Version

3.14

### OS and Platform

x86 Linux

AK2000:
## Summary

In the cloud exchange, unclosed redis connections are causing out  of resource errors. I think the issue is that people are still using version 0.3.1 of the code, which does not have a timeout on `listen`/`receive` requests. When this goes to the redis backend, the `blpop` call is made with no timeout, so it exists forever. Even when the client disconnects, no exception is raised within the handler, so the Redis connection persists until the app is restarted. This doesn't happen in the latest releases because the code by default has a timeout for `listen` requests. Still we need to deal with older requests (and requests that explicitly set the timeout to None). The tricky part is that these timeouts will raise unexpected exceptions in v0.3.1, so any agent that is expecting to live longer than that `timeout` without receiving requests must upgrade the newer version in order to use our hosted exchange.

## Related Issues

- Closes #386 

## Changes

## Testing

N/A

## Pull Request Checklist

- [] Docs have been updated and reviewed if relevant.

### DIFF
--- a/academy/exchange/cloud/app.py
+++ b/academy/exchange/cloud/app.py
@@ -330,6 +330,18 @@ async def _listen_mailbox_route(
         raw_mailbox_id,
     )
     timeout = data.get('timeout', request.app[TIMEOUT_KEY])
+    if timeout <= 0 or timeout > request.app[TIMEOUT_KEY]:
+        logger.warning(
+            f'Invalid timeout in listen request: {timeout}.',
+        )
+        return Response(
+            status=StatusCode.BAD_REQUEST.value,
+            text=(
+                'Invalid timeout (must be less than '
+                f'{request.app[TIMEOUT_KEY]} and greater than 0)'
+            ),
+        )
+
     client = get_client_info(request)
     status = await manager.check_mailbox(client, mailbox_id)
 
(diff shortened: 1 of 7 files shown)


benclifford:
> The tricky part is that these timeouts will raise unexpected exceptions in v0.3.1, so any agent that is expecting to live longer than that timeout without receiving requests must upgrade the newer version in order to use our hosted exchange.

That's got to happen eventually - no piece of the system should be allowed to allocate precious resources "forever" without constraint (e.g. timeout or credit card number).

And we're going to have way worse breaking changes with the hosted exchange I'm sure as other limits appear.

AK2000:
Marking as draft because there are some issues with the GlobusExchangeClient that will affect this PR (#391).

PR #397, opened 2026-04-20
Script for Running Agents

AK2000:
## Summary

Add a script to launch agents without the manager. The intention of this script is to be used for launching "persistent" agents, or agents that are used outside of the application/script that creates them. With this in mind, I created documentation illustrating how one might use the script to create a service. 

## Related Issues

- Related to #45
- Related to #389

## Changes

- [x] Documentation (no changes to the code)

## Testing

N/A

## Pull Request Checklist


AK2000:
In #389 there was discussion about changing the protocol not to use pickle. While I agree in spirit with that, we need pickle or a pickle like protocol at least for the `Agent` itself and the launch arguments. To me, this made any text-based protocol somewhat un-helpful, because it would still be combined in a file with binary.

gpauloski:
It's needed for the `Agent` and launch arguments for some fundamental reason, or just because we haven't created mechanisms for it?

I'm imagining you could allows users to implement a `from_toml` or similar classmethod on their `Agent`. And then they provide a config along the lines of `python -m academy.run --agent-spec /path/to/agent-spec.toml --exchange-spec /path/to/exchange-spec.toml`
```toml
[agent]
cls = "mypackage.agent.MyAgent"

[params]
foo = "bar"
spam = "eggs"
```
The `academy.run` module is responsible for importing the class using the provided fully qualified name and then initializing by passing the config to the classmethod.

You could do something similar for exchange things.
```toml
[exchange]
id = "<agent id>"
address = "..."
```

benclifford:
If you're imagining these files being created by humans, something like a yaml/toml/flavour of the day config file format is the way to go.

If you're imagining these files being created by code, "I'm imagining you could allows users to implement a from_toml or similar classmethod on their Agent" - or they could not have to implement such a method and instead use pickle!

It's probably pretty trivial to be able to read both?

I think both of the above cases are decent use cases, although I have a gut feeling that the human-curated version is the more serious one? In the machine launched case, you will come to admit (in a year or two) that there must always be a supervisor process, and we already support that without this PR.

docs/guides/persistent.md
```diff
@@ -0,0 +1,65 @@
...
+        spec = _RunSpec(
+            agent=MyAgentClass,
+            agent_args=<AgentArgs>,
+            agent_kwargs=<LaunchKwargs>,
+            exchange_factory=exchange_factory,
+            registration=registration,
+            config=RuntimeConfig(terminate_on_error=False),
+        )
+
+        with open('~/local/share/agents/MyAgentSpec.pkl', 'wb') as fp:
```
benclifford: this looks awful as a flow for the "sysadmin-curated" view of things -- i want my agent to start with a config file I can edit, not have to paste all this into a script just to edit a config file every time and hope I remembered it all.

docs/guides/persistent.md
```diff
@@ -0,0 +1,65 @@
...
+            pickle.dump(spec, fp)
+```
+
+Once we have created the registration and the spec, we can launch the agent as a separate [script][academy.run].
+```
+python -m academy.run --spec <path to spec file>
+```
+
+## Running an Agent with SystemD
```
benclifford: systemd is all lowercase

docs/guides/persistent.md
```diff
@@ -0,0 +1,65 @@
...
+
+By saving the `_RunSpec` as a file, we can set up a systemd service to keep the `Agent` alive. Note, these are general instructions for setting up an agent as a service, you may need to speak to your site administrator about system specific policies.
```
benclifford: we or you? (prefer you, this isn't a childrens book/academic article)

benclifford:
I think this is fine as a component of some bigger lifetime story.

This is fundamentally a "process" feature not a "python" feature and I'd be more comfortable with a process-oriented test that checks that a separate python process really can run an agent this way - although that's possibly a symptom of my disdain for mock based feature tests :laughing:

docs/guides/persistent.md
```diff
@@ -0,0 +1,65 @@
+# Building Persistent Agents
+
+Academy allows you to build agents as "micro-services" for scientific workflows --- the same agent can outlive a single application and can be reused in different contexts. in this guide, we walk through common techniques that you can use to deploy persistent agents and problems you may encounter.
```
benclifford: add a bit of example about how to talk to this agent once it has been created?

docs/guides/persistent.md
```diff
@@ -0,0 +1,65 @@
...
+
+!!! warning
+
+    This page is still under construction, more details and examples to come!
+
+## Running an Agent Without a Manager
+The most common (and recommended) way to run agentic applications with Academy is to use the [`Manager`][academy.manager.Manager] class. However, the manager class explicitly ties the lifetime of your agents to the lifetime of your manager.
+
+To create an agent without using a `manager`, we first need to create a persistent mailbox for the agent. We can use the `exchange_client.register_agent` functionality. Note, for persistent agents you must use a `ExchangeFactory` that lives outside of process memory, so the `RedisExchangeFactory`, `HybridExchangeFactory` or `HttpExchangeFactory`. We then save the `registration` with the arguments we are going to need to start the agent to a file using `pickle`.
```
benclifford: > Note, for persistent agents you must use a `ExchangeFactory` that lives outside of process memory, so the `RedisExchangeFactory`, `HybridExchangeFactory` or `HttpExchangeFactory`.

there's a slightly broader point that for distributed workflows, you need to use these kind of exchanges and more importantly, the entire collection of distributed workflows that is sharing agents needs to share *a single* exchange, right?

the current phrasing doesn't making it clear that it is a mistake to start up a  throwaway fresh redis exchange on my laptop, one-shot, to generate the .pkl file I later deploy on a supercomputer, which is an obvious thing to do.

sure you can escape the lifetime of a single manager, but you have to live within the lifetime of a single exchange.

that's more about running workflow(s) this way, rather than running the individual agents/scripts - so maybe this sentence gets rephrased and moves one level up?

docs/guides/persistent.md
benclifford: (which separately raises the question, am I allowed to fiddle with the parameters on each run of my agent?)

benclifford (dismissed):

benclifford:
> If you're imagining these files being created by humans, something like a yaml/toml/flavour of the day config file format is the way to go.

although this still needs the rest of the agent spec like the handle uuids etc, right? I think @AK2000 made that point and I missed it... that stuff doesn't make sense in yaml form as far as "why does it need to be visible/editable"

AK2000:
> It's needed for the Agent and launch arguments for some fundamental reason, or just because we haven't created mechanisms for it?
> I'm imagining you could allows users to implement a from_toml or similar classmethod on their Agent.

There is nothing fundamental. I guess you could have them implement a `from_toml` or something else, but to me, `pickle` was an easy way to offload work on to a trusted solution. I imagine for agents that take more than simple arguments, implementing this method gets a little cumbersome (i.e. you have to load types of arguments that may be beyond your code). Even for agents that only take simple arguments, I thought this created some unnecessary boiler plate that would be harder to use. 

> It's probably pretty trivial to be able to read both?

I agree, maybe we support both

> (which separately raises the question, am I allowed to fiddle with the parameters on each run of my agent?)

With this method you would not be able to.

---

I think my conclusion here is to break off more of the run config into a yaml file and put the agent as a separate config file. That way we can support the agent with pickle or (in the future) from toml.

AK2000:
I'm adding a note here about needing to import `AgentT` from `academy.identifier` when not type checking. When deserializing pydantic models, there's an issue where redefining `AgentT` causes issues trying to import the right type.

PR #398, opened 2026-04-24
Add `Manager.launch_batch()` for batched agent

GusEllerm:
## Summary

This PR improves the UX around user-facing auth prompts on the Globus Exchange. The underlying logic for batching auth was added in [a previous PR](https://github.com/academy-agents/academy/pull/387#issue-4257618820); this PR introduces the public API for it.

Before this PR, batching auth across a set of agents required driving `register_agents` and `launch` separately:

```
        factory=factory,
        log_config=recommended_logging(),
    ) as manager:
        regs: list[AgentRegistration[Any]] = await manager.register_agents(
            [
                (Greeter, None),
                (Shouter, None),
                (Coordinator, None),
            ],
        )

        greeter = await manager.launch(Greeter, registration=regs[0])
        shouter = await manager.launch(Shouter, registration=regs[1])
        coordinator = await manager.launch(
            Coordinator,
            args=(greeter, shouter),
            registration=regs[2],
        )
```

After this PR, the same flow uses `Manager.launch_batch()`:

```
        factory=factory,
        log_config=recommended_logging(),
    ) as manager:
        async with manager.launch_batch() as batch:
            greeter = await batch.queue(Greeter)
            shouter = await batch.queue(Shouter)
            coordinator = await batch.queue(
                Coordinator,
                args=(greeter, shouter),
            )
```

`batch.queue()` returns an unbound `Handle`; registration and launch happen on context-exit. On the Globus exchange, all queued agents are registered under a single consent prompt; other transports register sequentially. The batch can also be driven explicitly without a context manager via `await batch.submit()` -- see `examples/12-globus-exchange/run-12-explicit.py`.

- Builds on [#387](https://github.com/academy-agents/academy/pull/387#issue-4257618820)

## Changes

- [x] Documentation (no changes to the code)

## Testing

Unit coverage in `tests/unit/manager_test.py`, `tests/unit/handle_test.py`, and `tests/unit/exchange/client_test.py`:

## Pull Request Checklist

- [X] Relevant tags are added based on the types of changes.
- [X] Tests have been added to show the fix is effective or that the new feature works.
- [X] Docs have been updated and reviewed if relevant.

### DIFF
--- a/academy/handle.py
+++ b/academy/handle.py
@@ -95,6 +95,8 @@ class Handle(Generic[AgentT_co]):
         self._exchange = exchange
         self._registered_exchanges: WeakSet[ExchangeClient[Any]] = WeakSet()
         self.ignore_context = ignore_context
+        # Set by action/ping/shutdown; consulted by Manager._rebind_handle.
+        self._used_for_messaging = False
 
         if ignore_context and not exchange:
             raise ValueError(
@@ -208,6 +210,7 @@ class Handle(Generic[AgentT_co]):
                 (it self terminated or via another handle).
             Exception: Any exception raised by the action.
         """
+        self._used_for_messaging = True
         tag_id = uuid.uuid4()
         invocation_extra = {
             'academy.action': action,
@@ -331,6 +334,7 @@ class Handle(Generic[AgentT_co]):
                 (it self terminated or via another handle).
             TimeoutError: If the timeout is exceeded.
         """
+        self._used_for_messaging = True
         exchange = self.exchange
         self._register_with_exchange(exchange)
 
@@ -405,6 +409,7 @@ class Handle(Generic[AgentT_co]):
                 typically indicates the agent shutdown for another reason
                 (it self terminated or via another handle).
         """
+        self._used_for_messaging = True
         exchange = self.exchange
         self._register_with_exchange(exchange)
 
--- a/academy/handle.py
+++ b/academy/handle.py
@@ -74,7 +74,8 @@ class Handle(Generic[AgentT_co]):
         an exchange client.
 
     Args:
-        agent_id: ID of the remote agent.
+        agent_id: ID of the remote agent, or ``None`` to defer
+            binding (used by ``batch.queue()``).
         exchange: A default exchange client to be used if an exchange client
             is not configured in the current context.
         ignore_context: Ignore the current context and force use of `exchange`
@@ -86,17 +87,15 @@ class Handle(Generic[AgentT_co]):
 
     def __init__(
         self,
-        agent_id: AgentId[AgentT_co],
+        agent_id: AgentId[AgentT_co] | None = None,
         *,
         exchange: ExchangeClient[Any] | None = None,
         ignore_context: bool = False,
     ) -> None:
-        self.agent_id = agent_id
+        self._agent_id: AgentId[AgentT_co] | None = agent_id
         self._exchange = exchange
         self._registered_exchanges: WeakSet[ExchangeClient[Any]] = WeakSet()
         self.ignore_context = ignore_context
-        # Set by action/ping/shutdown; consulted by Manager._rebind_handle.
-        self._used_for_messaging = False
 
         if ignore_context and not exchange:
             raise ValueError(
@@ -114,6 +113,25 @@ class Handle(Generic[AgentT_co]):
 
         if self._exchange is not None:
             self._register_with_exchange(self._exchange)
+
+    @property
+    def agent_id(self) -> AgentId[AgentT_co]:
+        """ID of the remote agent.
+
+        Raises:
+            RuntimeError: If the handle is not bound.
+        """
+        if self._agent_id is None:
+            raise RuntimeError(
+                'Handle is not bound to a registered agent. '
+                'Submit the enclosing batch before reading '
+                'agent_id.',
+            )
+        return self._agent_id
+
+    @agent_id.setter
+    def agent_id(self, value: AgentId[AgentT_co]) -> None:
+        self._agent_id = value
 
     @property
     def exchange(self) -> ExchangeClient[Any]:
@@ -148,18 +166,20 @@ class Handle(Generic[AgentT_co]):
             raise PicklingError(
                 'Handle with ignore_context=True is not pickle-able',
             )
-        return (Handle, (self.agent_id,))
+        if self._agent_id is None:
+            raise PicklingError('Cannot pickle an unbound handle.')
+        return (Handle, (self._agent_id,))
 
     def __repr__(self) -> str:
         return (
-            f'{type(self).__name__}(agent_id={self.agent_id!r}, '
+            f'{type(self).__name__}(agent_id={self._agent_id!r}, '
             f'exchange={self._exchange!r}, '
             f'ignore_context={self.ignore_context!r})'
         )
 
     def __str__(self) -> str:
         name = type(self).__name__
-        return f'{name}<agent: {self.agent_id}>'
+        return f'{name}<agent: {self._agent_id}>'
 
     def __getattr__(self, name: str) -> Any:
         async def remote_method_call(*args: Any, **kwargs: Any) -> R:
@@ -210,7 +230,6 @@ class Handle(Generic[AgentT_co]):
                 (it self terminated or via another handle).
             Exception: Any exception raised by the action.
         """
-        self._used_for_messaging = True
         tag_id = uuid.uuid4()
         invocation_extra = {
             'academy.action': action,
@@ -334,7 +353,6 @@ class Handle(Generic[AgentT_co]):
                 (it self terminated or via another handle).
             TimeoutError: If the timeout is exceeded.
         """
-        self._used_for_messaging = True
         exchange = self.exchange
         self._register_with_exchange(exchange)
 
@@ -409,7 +427,6 @@ class Handle(Generic[AgentT_co]):
                 typically indicates the agent shutdown for another reason
                 (it self terminated or via another handle).
         """
-        self._used_for_messaging = True
         exchange = self.exchange
         self._register_with_exchange(exchange)
 
--- a/academy/handle.py
+++ b/academy/handle.py
@@ -131,6 +131,7 @@ class Handle(Generic[AgentT_co]):
 
     @agent_id.setter
     def agent_id(self, value: AgentId[AgentT_co]) -> None:
+        assert self._agent_id is None, 'Handle is already bound.'
         self._agent_id = value
 
     @property
@@ -463,7 +464,7 @@ class ProxyHandle(Handle[AgentT]):
 
     def __init__(self, agent: AgentT) -> None:
         self.agent = agent
-        self.agent_id: AgentId[AgentT] = AgentId.new()
+        self._agent_id: AgentId[AgentT] = AgentId.new()
         self._agent_closed = False
 
     def __repr__(self) -> str:
--- a/academy/manager.py
+++ b/academy/manager.py
@@ -275,8 +275,7 @@ class _BatchLauncher:
             orphan_registrations = registrations[launch_index:]
             handles = self._manager._handles
             for intent in orphan_intents:
-                if handles.get(intent.handle.agent_id) is intent.handle:
-                    del handles[intent.handle.agent_id]
+                handles.pop(intent.handle.agent_id, None)
             await self._terminate_orphans(orphan_registrations)
             raise
 
@@ -288,9 +287,6 @@ class _BatchLauncher:
 
         The original launch exception must not be masked.
         """
-        if not orphans:
-            return
-
         exchange = self._manager.exchange_client
         results = await asyncio.gather(
             *(exchange.terminate(reg.agent_id) for reg in orphans),
@@ -785,7 +781,7 @@ class Manager(Generic[ExchangeTransportT], NoPickleMixin):
                 f'Cannot rebind handle to {new_agent_id}: cache '
                 f'already holds a different handle for that ID.',
             )
-        if self._handles.get(old_agent_id) is handle:
+        if self._handles.get(old_agent_id) is handle:  # pragma: no branch
             del self._handles[old_agent_id]
         handle.agent_id = new_agent_id
         self._handles[new_agent_id] = handle
--- a/academy/manager.py
+++ b/academy/manager.py
@@ -758,29 +758,20 @@ class Manager(Generic[ExchangeTransportT], NoPickleMixin):
         be the real ID and this method (plus the
         ``_used_for_messaging`` guard on :class:`Handle`) could go
         away. Deferred because it changes the transport protocol.
-
-        Raises:
-            RuntimeError: If ``handle`` has already been used for
-                messaging, or if ``new_agent_id`` is already a cache
-                key for a different handle. Uses ``RuntimeError``
-                rather than ``assert`` so the check survives
-                ``python -O``.
         """
         old_agent_id = handle.agent_id
         if old_agent_id == new_agent_id:
             return
-        if getattr(handle, '_used_for_messaging', False):
-            raise RuntimeError(
-                f'Cannot rebind handle {old_agent_id}: already used '
-                'for messaging. Pass the Handle itself (not '
-                'handle.agent_id) as a sibling launch_batch argument.',
-            )
+        assert not handle._used_for_messaging, (
+            f'Cannot rebind handle {old_agent_id}: already used for '
+            'messaging. Pass the Handle itself (not handle.agent_id) '
+            'as a sibling launch_batch argument.'
+        )
         existing = self._handles.get(new_agent_id)
-        if existing is not None and existing is not handle:
-            raise RuntimeError(
-                f'Cannot rebind handle to {new_agent_id}: cache '
-                f'already holds a different handle for that ID.',
-            )
+        assert existing is None or existing is handle, (
+            f'Cannot rebind handle to {new_agent_id}: cache already '
+            'holds a different handle for that ID.'
+        )
         if self._handles.get(old_agent_id) is handle:  # pragma: no branch
             del self._handles[old_agent_id]
         handle.agent_id = new_agent_id
--- a/academy/manager.py
+++ b/academy/manager.py
@@ -100,6 +100,205 @@ class _ACB(Generic[AgentT]):
     task: asyncio.Task[None]
 
 
+@dataclasses.dataclass
+class _LaunchIntent(Generic[AgentT]):
+    # Fields mirror Manager.launch's parameters; a dict[str, Any]
+    # would drop static typing.
+    agent: AgentT | type[AgentT]
+    handle: Handle[AgentT]
+    name: str | None = None
+    args: tuple[Any, ...] | None = None
+    kwargs: dict[str, Any] | None = None
+    config: RuntimeConfig | None = None
+    executor: str | None = None
+    submit_kwargs: dict[str, Any] | None = None
+    log_config: LogConfig | None = None
+
+
+class _BatchLauncher:
+    """Context manager that batches multiple agent launches.
+
+    Returned by :meth:`Manager.launch_batch`. On context exit
+    without error, queued agents are registered together (one
+    auth consent prompt on transports that batch auth) and launched.
+
+    Args:
+        manager: Manager that owns this batch.
+    """
+
+    def __init__(self, manager: Manager[Any]) -> None:
+        self._manager = manager
+        self._intents: list[_LaunchIntent[Any]] = []
+        self._closed = False
+
+    async def __aenter__(self) -> Self:
+        return self
+
+    async def __aexit__(
+        self,
+        exc_type: type[BaseException] | None,
+        exc_value: BaseException | None,
+        exc_traceback: TracebackType | None,
+    ) -> None:
+        self._closed = True
+        if exc_value is not None:
+            # Intents are retained for caller inspection on this path.
+            self._evict_placeholder_handles()
+            return
+        if not self._intents:
+            return
+        try:
+            await self._commit()
+        finally:
+            # Drop per-intent references; the batch object may outlive
+            # ``async with`` and the manager's caches own the state now.
+            self._intents.clear()
+
+    def _evict_placeholder_handles(self) -> None:
+        """Remove queued intents' handles from the manager cache."""
+        handles = self._manager._handles
+        for intent in self._intents:
+            if handles.get(intent.handle.agent_id) is intent.handle:
+                del handles[intent.handle.agent_id]
+
+    async def launch(  # noqa: PLR0913
+        self,
+        agent: AgentT | type[AgentT],
+        *,
+        args: tuple[Any, ...] | None = None,
+        kwargs: dict[str, Any] | None = None,
+        config: RuntimeConfig | None = None,
+        executor: str | None = None,
+        submit_kwargs: dict[str, Any] | None = None,
+        name: str | None = None,
+        log_config: LogConfig | None = None,
+    ) -> Handle[AgentT]:
+        """Queue an agent launch. Committed on context exit.
+
+        Parameters mirror :meth:`Manager.launch` exactly except for
+        ``registration`` (the batch manages registration itself).
+
+        Returns:
+            Handle to the agent. Usable as a sibling ``batch.launch``
+            argument; not messageable until the batch commits.
+
+        Warning:
+            Pass the :class:`Handle` itself to sibling launches,
+            not ``handle.agent_id``. The handle is rebound in place
+            at commit time; an ``agent_id`` captured into ``args``
+            before commit is a stale placeholder UUID.
+
+        Raises:
+            RuntimeError: If the batch has already exited.
+        """
+        if self._closed:
+            raise RuntimeError('batch is closed')
+
+        # Pre-mint so the returned handle can be a sibling-launch arg.
+        aid: AgentId[AgentT] = AgentId.new(name=name)
+        handle = self._manager.get_handle(aid)
+
+        self._intents.append(
+            _LaunchIntent(
+                agent=agent,
+                handle=handle,
+                name=name,
+                args=args,
+                kwargs=kwargs,
+                config=config,
+                executor=executor,
+                submit_kwargs=submit_kwargs,
+                log_config=log_config,
+            ),
+        )
+        return handle
+
+    async def _commit(self) -> None:
+        """Register queued intents together, then launch each.
+
+        On transports that implement
+        :meth:`ExchangeTransport.register_agents` (e.g. Globus), this
+        produces one auth consent prompt for the whole batch; other
+        transports fall back to sequential registration. Each
+        pre-minted handle is then rebound to the transport-minted ID
+        before launch.
+        """
+        specs: list[tuple[type[Any], str | None]] = [
+            (
+                intent.agent
+                if isinstance(intent.agent, type)
+                else type(intent.agent),
+                intent.name,
+            )
+            for intent in self._intents
+        ]
+        registrations = await self._manager.register_agents(specs)
+
+        launch_index = 0
+        try:
+            for intent, registration in zip(
+                self._intents,
+                registrations,
+                strict=True,
+            ):
+                self._manager._rebind_handle(
+                    intent.handle,
+                    registration.agent_id,
+                )
+                await self._manager.launch(
+                    intent.agent,
+                    args=intent.args,
+                    kwargs=intent.kwargs,
+                    config=intent.config,
+                    executor=intent.executor,
+                    submit_kwargs=intent.submit_kwargs,
+                    name=intent.name,
+                    registration=registration,
+                    log_config=intent.log_config,
+                )
+                launch_index += 1
+        except Exception:
+            # CancelledError deliberately excluded: registrations past
+            # ``launch_index`` will leak exchange-side, but awaiting
+            # cleanup inside a cancellation frame risks a hung shutdown.
+            orphan_intents = self._intents[launch_index:]
+            orphan_registrations = registrations[launch_index:]
+            handles = self._manager._handles
+            for intent in orphan_intents:
+                if handles.get(intent.handle.agent_id) is intent.handle:
+                    del handles[intent.handle.agent_id]
+            await self._terminate_orphans(orphan_registrations)
+            raise
+
+    async def _terminate_orphans(
+        self,
+        orphans: list[AgentRegistration[Any]],
+    ) -> None:
+        """Best-effort parallel termination; failures logged, never raised.
+
+        The original launch exception must not be masked.
+        """
+        if not orphans:
+            return
+
+        exchange = self._manager.exchange_client
+        results = await asyncio.gather(
+            *(exchange.terminate(reg.agent_id) for reg in orphans),
+            return_exceptions=True,
+        )
+        for registration, result in zip(orphans, results, strict=True):
+            if isinstance(result, Exception):
+                logger.error(
+                    'Failed to terminate orphan registration %s during '
+                    'launch_batch commit rollback.',
+                    registration.agent_id,
+                    exc_info=result,
+                    extra={
+                        'academy.agent_id': registration.agent_id,
+                    },
+                )
+
+
 class Manager(Generic[ExchangeTransportT], NoPickleMixin):
     """Launch and manage running agents.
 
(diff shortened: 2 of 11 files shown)


academy/manager.py
```diff
@@ -100,6 +100,205 @@ class _ACB(Generic[AgentT]):
...
+            self._evict_placeholder_handles()
+            return
+        if not self._intents:
+            return
+        try:
+            await self._commit()
+        finally:
+            # Drop per-intent references; the batch object may outlive
+            # ``async with`` and the manager's caches own the state now.
+            self._intents.clear()
```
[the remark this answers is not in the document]
GusEllerm: When you queue an agent with `batch.launch()`, a Handle is returned and stashed in the manager's cache under a placeholder ID. If the batch auth fails before the placeholder gets swapped with the real ID (from the exchange) the placeholder entry was orphaned. The fix is 1. remember the placeholder on the intent so we can find it later, 2. on failure walk the queued intents and remove their entires from the cache, 3. clear the intent list only after cleanup runs.

GusEllerm:
The `_rebind_handle` TODO exists because I wanted to maintain the transport as the only place where IDs are minted. If the `register_agents` function accepted pre-minted IDs from the caller I could simplify the logic here, but that would mean a transport-protocol change which probs should not happen before a discussion. 

There are also questions around context managers all the way down and the UX decisions behind that.

docs/migration.md
```diff
@@ -29,6 +29,19 @@ Multiple log contexts can be active in a process at any one time - for example,
 
 Academy comes with three ways of configuring logging: to the console, to a log file, and to a shared home directory JSON logfile store. Developers can implement new logging configurations by subclassing the LogConfig class.
 
+#### New: `Manager.launch_batch()`
+
+Small ergonomics addition for the Globus transport: launches queued inside a `launch_batch()` block are registered under a single auth consent prompt instead of one per agent. Other transports behave identically to separate `Manager.launch()` calls.
```
benclifford: this is not small

benclifford:
what happens when a user tries to send a message using a magic quasihandles?

academy/manager.py
```diff
@@ -530,6 +739,84 @@ def get_handle(
...
+        if old_agent_id == new_agent_id:
+            return
+        if getattr(handle, '_used_for_messaging', False):
+            raise RuntimeError(
+                f'Cannot rebind handle {old_agent_id}: already used '
+                'for messaging. Pass the Handle itself (not '
+                'handle.agent_id) as a sibling launch_batch argument.',
+            )
+        existing = self._handles.get(new_agent_id)
+        if existing is not None and existing is not handle:
```
benclifford: are you expecting this `if` to fire and/or the `if` below to fire in the normal course of operation? (aka are you expecting there to be multiple rebinds if the rest of academy is correctly implemented?)

GusEllerm: No -- these are pure defensive guards for handles pointing to the wrong agents, but that should never happen. Do we prefer assuming robust operation?

GusEllerm:
Re `what happens when a user tries to send a message using a magic quasihandles?`

The placeholder `AgentId` isn't registered until `__aexit__` commits, so you get a `BadEntityIdError` immediately. `Handle.action`/`ping`/`shutdown` sets [`_used_for_messaging = True`](https://github.com/academy-agents/academy/pull/398/changes#diff-9f354fa0a6a96fd1ef5fa35e36dacc76971fe85e591bbefcd21cbd1639e3dfbdR213) before awaiting send, so at commit `_rebind_handle` refuses to silently rebind, raising a [`RuntimeError`](https://github.com/academy-agents/academy/pull/398/changes#diff-9a92aeccb1667d55f31ee880f4c051f6df7dc958a235e4db7d388d5490716a75R772) pointing to the correct API.

academy/manager.py
benclifford: I think its good to have checks on internal state like this. I think they would be better as asserts, if they really are assertions about how the internal state should be. The comment about "users might turn off asserts so I MEGA-ASSERT()" in the docstring is a bit of a red herring: the code is broken and turning off asserts "merely" changes how that breakage appears, and that's part of the usage contract of turning off asserts.

benclifford: (i mean the code is broken if the assert fires or if the assert-would-fire-but-was-disabled, not that it is broken in this PR)

academy/manager.py
```diff
@@ -100,6 +100,211 @@ class _ACB(Generic[AgentT]):
...
+                    executor=intent.executor,
+                    submit_kwargs=intent.submit_kwargs,
+                    name=intent.name,
+                    registration=registration,
+                    log_config=intent.log_config,
+                )
+                launch_index += 1
+        except Exception:
+            # CancelledError deliberately excluded: registrations past
+            # ``launch_index`` will leak exchange-side, but awaiting
```
benclifford: I think this isn't seriously worse than other registrations on the exchange that aren't don't go through the full life-cycle, which is going to happen in sufficiently crashing environments where cleanup doesn't happen.

academy/handle.py
```diff
@@ -95,6 +95,8 @@ def __init__(
         self._exchange = exchange
         self._registered_exchanges: WeakSet[ExchangeClient[Any]] = WeakSet()
         self.ignore_context = ignore_context
+        # Set by action/ping/shutdown; consulted by Manager._rebind_handle.
+        self._used_for_messaging = False
```
benclifford: thinking about the representation of states.

in the beginning, this is set to false, and there is a fake but indistinguishable from valid agent ID in the handle.

in the end, this is set to true, and there is a different, now valid but still indistinguishable from fake agent ID in the handle.

_used_for_messaging is almost but not quite a flag of whether the agent ID is valid, I think - because I think this can be set by a user accidentally doing something with the handle inside a launch batch context?

So then, should this flag actually be for if the fake agent ID has been replaced with a real agent ID?

And then if so, should the agent then become `Optional` and that's the flag?

docs/migration.md
```diff
@@ -122,57 +140,56 @@ Agents are now derived from [`Agent`][academy.agent.Agent] (previously, `Behavio
 
 Summary:
 
-* `academy.agent.Agent` is renamed [`academy.runtime.Runtime`][academy.runtime.Runtime].
-* `academy.behavior.Behavior` is renamed [`academy.agent.Agent`][academy.agent.Agent].
-* `academy.identifier.ClientId` is renamed [`academy.identifier.UserId`][academy.identifier.UserId].
+- `academy.agent.Agent` is renamed [`academy.runtime.Runtime`][academy.runtime.Runtime].
```
AK2000: Minor thing, but unless there is a reason to make this change, I don't think we need to change '*' to '-', and I would prefer to leave it just to reduce churn

academy/manager.py
```diff
@@ -530,6 +739,84 @@ def get_handle(
...
+            RuntimeError: If ``handle`` has already been used for
+                messaging, or if ``new_agent_id`` is already a cache
+                key for a different handle. Uses ``RuntimeError``
+                rather than ``assert`` so the check survives
+                ``python -O``.
+        """
+        old_agent_id = handle.agent_id
+        if old_agent_id == new_agent_id:
+            return
+        if getattr(handle, '_used_for_messaging', False):
```
AK2000: Why do we need to use `getattr` here? Is there a code path that I'm missing where the `_used_for_messaging` attribute won't exist?

AK2000 (dismissed):
Looks good to me, but I agree with Ben that the flag `_used_for_messaging` is odd. What if instead we had `_BatchLauncher.launch` return a `Handle(agent_id=None)` that would raise a runtime error if called. This `Handle` would not correspond to any entry in `manager._handles`. And then we could do the "rebinding" solely in `_BatchLauncher._commit` instead of needing a new method in the manager to delete the old entry from the dictionary and create the new handle.

academy/manager.py
GusEllerm: We don't. Fixed.

academy/manager.py
GusEllerm: Understood, other failure modes will orphan registrations. This is not a unique failure mode. Do you think highlighting this in the comment puts too much emphasis on that possibility?

academy/handle.py
GusEllerm: Yep, no guarantee .action() is called post-commit. A user who tries messaging inside the batch context misroutes to a non-existent mailbox.
                                                                                      
@AK2000 's `Optional[AgentId]` suggestion could work: making `agent_id` optional and binding in `_commit` removes alot of logic from this PR. Pre-commit messaging fails at the call site rather than silently misrouting.                               

The cost would be propagating `Optional` through Handle's read sites, plus any external caller reading handle.agent_id, would all need to handle **None**. 

The TODO on `_rebind_handle` points at another refactor (transports accept caller-minted IDs) that achieves the same effect without the `Optional`, but that's a transport protocol change and probs out of scope here.                                                            

Happy to fold this in if you both think the read None propagation is worth it. @benclifford @AK2000

academy/manager.py
benclifford: i think the comment is fine. i just made my comment so you didn't feel guilty.

academy/handle.py
benclifford: > The cost would be propagating Optional through Handle's read sites

Some check like this should probably be happening one way or another to stop users sending stuff to the handle before it is properly bound - because it's hard to debug mystery behaviour otherwise.

> The TODO on _rebind_handle points at another refactor (transports accept caller-minted IDs) 

This is probably the right way to go eventually - following the idea that the ID of an entity should be generated at creation time.

I don't have a feel in my head for how much code is reading agent IDs out of handles, to know if it is hard. But this PR introduces the notion that reading agent IDs out of handles is now not always valid - whether you make it explicit by this type mechanism, or if you make it implicit by letting a user read a fake value and then laugh at them spending a week debugging it.

So I think you should use optional.

AK2000: What about something like this:
```
class PossibilyIncompleteHandle(Handle[AgentT]):
    """Handle that does not point to a registered agent.

    A PossiblyIncompleteHandle is a handle that is created before
    a mailbox is created in the exchange, for instance because
    the handle was created during a batch. If the registration
    completes and this handle is updated to point to the agent
    id, then the Handle behaves like a regular Handle.
    """

    def __init__(
        self,
        agent_id: AgentId[AgentT_co],
        *,
        ignore_context: bool = False,
    ) -> None:
        super().__init__(
            None,  # type: ignore[arg-type]
            exchange=exchange,
            ignore_context=ignore_context,
        )
        self._agent_id = None

    @property
    def agent_id(self) -> AgentId[AgentT]:
        """Return agent_id or raise error if agent_id has not been set."""
        if self._agent_id is None:
            raise RuntimeError(
                'Handle points to unregistered agent. This is caused by'
                'a unfinished agent registration.',
            )
        return self._agent_id

    @agent_id.setter
    def agent_id(self, agent_id: AgentId[AgentT]) -> None:
        """Set agent id."""
        self._agent_id = agent_id
```
That would avoid the need for propagating the `None` type to any uses of the handle because we do the check here.

benclifford: if you're prepared to have handle instances that raise exceptions when you look at their agent_id, you could implement that property on Handle without another subclass - the subclass doesn't bring much/anything, I think.

If you're making a type check to see if it is this specific subclass, then you could also make a check for `None` at the same point?

I thought about suggesting that but didn't get round to it, and I am probably supportive of the idea if it turns out there are so many agent_id references that you don't want to deal with `None` values.

(the typing is probably the other way round: there's a class of maybehandles that might have an agent ID, and a subclass of that is the type of handles that definitely have an ID - but this is all just mirroring `Optional` in complicated ways that hide that fact that you have to deal with the fact that this PR #398 introduces handles-with-out-agents *somewhere*)

benclifford: (bracketed comment is basically the same as saying `Optional[AgentId]` is a supertype, not a subtype, of `AgentId`)

I probably prefer that we admit this is observable to a user and return `None` and perhaps be forced to acknowledge that if you have a handle and are trying to get the agent ID, you're doing something wrong.

*or* go make the protocol changes that @GusEllerm talked about! (which will bring its own problems of state, I think :laughing: )

academy/manager.py
```diff
@@ -99,6 +99,211 @@ class _ACB(Generic[AgentT]):
...
+    config: RuntimeConfig | None = None
+    executor: str | None = None
+    submit_kwargs: dict[str, Any] | None = None
+    log_config: LogConfig | None = None
+
+
+class _BatchLauncher:
+    """Context manager that batches multiple agent launches.
+
+    Returned by :meth:`Manager.launch_batch`. On context exit
```
AK2000: The :meth: doc comments won't be parsed correctly. Sorry I didn't notice this previously

academy/manager.py
```diff
@@ -99,6 +99,211 @@ class _ACB(Generic[AgentT]):
...
+
+    def __init__(self, manager: Manager[Any]) -> None:
+        self._manager = manager
+        self._intents: list[_LaunchIntent[Any]] = []
+        self._closed = False
+
+    async def __aenter__(self) -> Self:
+        return self
+
+    async def __aexit__(
```
AK2000, line 135: Can we have methods that allow this to be used as an object without a context manager too? I think that should mean simplifying this method a little and making `commit` public

academy/manager.py
```diff
@@ -99,6 +99,211 @@ class _ACB(Generic[AgentT]):
...
+
+        Raises:
+            RuntimeError: If the batch has already exited.
+        """
+        if self._closed:
+            raise RuntimeError('batch is closed')
+
+        # Pre-mint so the returned handle can be a sibling-launch arg.
+        aid: AgentId[AgentT] = AgentId.new(name=name)
+        handle = self._manager.get_handle(aid)
```
AK2000: I don't think we should add these fake handles to manager._handles --- this is complicating the design quite a bit. If we don't do this, we don't have to go back and edit the dictionary, and we don't have to evict handles and other complications.

academy/manager.py
```diff
@@ -99,6 +99,211 @@ class _ACB(Generic[AgentT]):
...
+
+        The original launch exception must not be masked.
+        """
+        exchange = self._manager.exchange_client
+        results = await asyncio.gather(
+            *(exchange.terminate(reg.agent_id) for reg in orphans),
+            return_exceptions=True,
+        )
+        for registration, result in zip(orphans, results, strict=True):
+            if isinstance(result, Exception):
```
AK2000: I don't think that we should suppress exceptions from cancellations. I think it's good to produce error traces that say things like "Exception during handling of the previous exception" when there are multiple errors that we don't expect, and there is no reason to  expect any `Exception` after one launch fails.

academy/manager.py
```diff
@@ -434,6 +639,10 @@ async def launch(  # noqa: PLR0913
     ) -> Handle[AgentT]:
         """Launch a new agent with a specified agent.
 
+        Note:
+            Parameters are mirrored in :class:`_LaunchIntent` and
```
AK2000: Same thing about these doc strings --- the :class: and :method: won't be parsed.

tests/unit/exchange/client_test.py
```diff
@@ -117,6 +117,36 @@ async def test_register_agents_empty(
     assert registrations == []
 
 
+@pytest.mark.asyncio
+async def test_register_agents_fallback_rolls_back(
+    client: UserExchangeClient[Any],
+    monkeypatch: pytest.MonkeyPatch,
+    caplog: pytest.LogCaptureFixture,
+) -> None:
+    monkeypatch.delattr(
```
AK2000: Can we use `mock.patch.object` instead of `monkeypatch`, then we can also assert that this was called.

tests/unit/exchange/client_test.py
```diff
@@ -117,6 +117,36 @@ async def test_register_agents_empty(
...
+        'register_agents',
+        raising=False,
+    )
+    first = await client.register_agent(EmptyAgent)
+    second = await client.register_agent(EmptyAgent)
+    register = mock.AsyncMock(
+        side_effect=[first, second, RuntimeError('boom')],
+    )
+    terminate = mock.AsyncMock(side_effect=[None, RuntimeError('t-fail')])
+    monkeypatch.setattr(client, 'register_agent', register)
```
AK2000: Same here --- `mock.patch` will replace the attribute with a mock automatically and be more consistent with the other tests

AK2000 (changes requested):
I went through and found some other things that I hadn't caught the first time.

tests/unit/exchange/client_test.py
AK2000: I think this should be true most places where monkeypatch is used

academy/manager.py
AK2000: Sorry one other thought. `commit` might imply the operations are somehow atomic (i.e. all the agents are launched or none of them are). Maybe something like `queue` to add agents to a batch and `submit` to finalize the batch? Open to other suggestions too

GusEllerm:
Pushing updates to address reviews.

[`Handle.agent_id`](https://github.com/academy-agents/academy/pull/398/changes#diff-9f354fa0a6a96fd1ef5fa35e36dacc76971fe85e591bbefcd21cbd1639e3dfbdR117-R135) is now a property that raises when the handle isn't bound. Subsequently much of the logic around handling un-bound handles has been dropped. There are no fake-but-maybe-valid ids to track. `manager._handles` only holds bound handles.

The batch API has different semantics. `batch.launch` -> [`batch.queue`](https://github.com/academy-agents/academy/pull/398/changes#diff-9a92aeccb1667d55f31ee880f4c051f6df7dc958a235e4db7d388d5490716a75R148-R192) (calling queue enqueues an intent, it does not launch an agent). `_commit` is replaced by public [`submit()`](https://github.com/academy-agents/academy/pull/398/changes#diff-9a92aeccb1667d55f31ee880f4c051f6df7dc958a235e4db7d388d5490716a75R194-R265) and [`abort()`](https://github.com/academy-agents/academy/pull/398/changes#diff-9a92aeccb1667d55f31ee880f4c051f6df7dc958a235e4db7d388d5490716a75R267-R272). [`__aexit__`](https://github.com/academy-agents/academy/pull/398/changes#diff-9a92aeccb1667d55f31ee880f4c051f6df7dc958a235e4db7d388d5490716a75R135-R146) is now a short dispatcher. An example of the non-context manager flow for batch auth can be found in the [new example](https://github.com/academy-agents/academy/pull/398/changes#diff-cc3da0be472dde4eac62cec80ed867010ae802537bfb92a332a822f5bc52a552R67-R94). 

[Tests migrated](https://github.com/academy-agents/academy/pull/398/commits/138122e3fff9698e9156faf2cae58f5828364aaa) from `monkeypatch` to `mock.patch.object`. [Sphinx syntax dropped](https://github.com/academy-agents/academy/pull/398/commits/e35199b4d629512d336626e73c638e00a99454d4) in favour of mkdocs (note this also affects the [earlier PR](https://github.com/academy-agents/academy/pull/387#issue-4257618820) for batching)

academy/manager.py
```diff
@@ -99,6 +99,204 @@ class _ACB(Generic[AgentT]):
...
+
+            launch_index = 0
+            try:
+                for intent, registration in zip(
+                    self._intents,
+                    registrations,
+                    strict=True,
+                ):
+                    intent.handle.agent_id = registration.agent_id
+                    self._manager._handles[registration.agent_id] = (
```
AK2000: Manager.launch will take care of this, we shouldn't need to do this

academy/manager.py
AK2000: I still don't think we should be surpressing this error, if we just call gather without return_exceptions, these exceptions will be raised to the user. `gather` won't cancel the other tasks if one returns an exception, which is the behavior we want.

academy/manager.py
```diff
@@ -434,6 +632,10 @@ async def launch(  # noqa: PLR0913
     ) -> Handle[AgentT]:
         """Launch a new agent with a specified agent.
 
+        Note:
+            Parameters are mirrored in `_LaunchIntent` and
+            `_BatchLauncher.launch`; keep all three in sync.
```
AK2000: Minor: "Parameters are mirrored in `_BatchLauncher.queue`

academy/manager.py
```diff
@@ -528,6 +730,39 @@ def get_handle(
...
+        every agent queued inside the ``async with`` block is
+        registered under a single auth consent prompt. Other
+        transports fall back to sequential registration.
+
+        Handles returned by ``batch.queue`` can be passed as
+        arguments to sibling ``batch.queue`` calls; they are not
+        messageable until the batch is submitted on exit.
+
+        Example:
+            .. code-block:: python
```
AK2000: Can we change this to `mkdocs` style?

academy/manager.py
```diff
@@ -99,6 +99,204 @@ class _ACB(Generic[AgentT]):
...
+                handles = self._manager._handles
+                for intent in orphan_intents:
+                    if intent.handle._agent_id is not None:
+                        handles.pop(intent.handle._agent_id, None)
+                await self._terminate_orphans(orphan_registrations)
+                raise
+        finally:
+            self._intents.clear()
+
+    def abort(self) -> None:
```
AK2000: I don't even think we need this method. Aborting can just be not calling `submit`, right?

examples/12-globus-exchange/run-12-explicit.py
```diff
@@ -0,0 +1,98 @@
...
+        batch = manager.launch_batch()
+        try:
+            greeter = await batch.queue(Greeter)
+            shouter = await batch.queue(Shouter)
+            coordinator = await batch.queue(
+                Coordinator,
+                args=(greeter, shouter),
+            )
+            await batch.submit()
+        except Exception:
```
AK2000: If we get rid of abort, we can simplify this example, and abort doesn't really do any necessary clean up.

examples/12-globus-exchange/run-12-explicit.py
```diff
@@ -0,0 +1,98 @@
+"""Explicit-driving variant of run-12.py.
```
AK2000, line 1: In addition to adding this in an example can you add documentation in docs/guides/persistant.md

academy/manager.py
```diff
@@ -99,6 +99,204 @@ class _ACB(Generic[AgentT]):
...
+                    launch_index += 1
+            except Exception:
+                # CancelledError deliberately excluded: registrations
+                # past ``launch_index`` will leak exchange-side, but
+                # awaiting cleanup inside a cancellation frame risks a
+                # hung shutdown.
+                orphan_intents = self._intents[launch_index:]
+                orphan_registrations = registrations[launch_index:]
+                handles = self._manager._handles
+                for intent in orphan_intents:
```
AK2000: If I track the code correctly, there is only ever exactly 1 orphaned handle, that triggers the condition on line 260, the one that was currently being launched. Can we be more explicit about that clean up here instead of iterating through all remaining intents?

tests/unit/manager_test.py
```diff
@@ -162,6 +170,472 @@ async def test_register_agents_and_launch(
...
+        assert batch._intents[2].agent is EmptyAgent
+        assert batch._intents[0].name == 'a'
+        assert batch._intents[1].name == 'b'
+        assert batch._intents[2].name == 'c'
+    ids = {h1.agent_id, h2.agent_id, h3.agent_id}
+    assert len(ids) == 3  # noqa: PLR2004
+
+
+@pytest.mark.asyncio
+async def test_launch_batch_handle_usable_as_sibling_arg(
```
AK2000: I think it would be more prudent to test if the sibling agent id is propagated after launch (i.e. the sibling does not see a unbound handle)

tests/unit/manager_test.py
```diff
@@ -162,6 +170,472 @@ async def test_register_agents_and_launch(
...
+    manager: Manager[LocalExchangeTransport],
+) -> None:
+    async with manager.launch_batch() as batch:
+        handle = await batch.queue(EmptyAgent, name='worker-1')
+        assert batch._intents[0].name == 'worker-1'
+    assert handle.agent_id.name == 'worker-1'
+
+
+@pytest.mark.asyncio
+async def test_launch_batch_multiple_launches_preserve_order(
```
AK2000: I don't know if this is testing anything useful. If this is a Claude test can we just remove it?

tests/unit/manager_test.py
```diff
@@ -162,6 +170,472 @@ async def test_register_agents_and_launch(
...
+    assert len(manager.running()) == 2  # noqa: PLR2004
+    assert manager._handles[h1.agent_id] is h1
+    assert manager._handles[h2.agent_id] is h2
+    running_ids = manager.running()
+    assert h1.agent_id in running_ids
+    assert h2.agent_id in running_ids
+
+
+@pytest.mark.asyncio
+async def test_launch_batch_submit_binds_handle(
```
AK2000: I don't think this tests anything not covered by the previous test. Can we remove it?

tests/unit/manager_test.py
```diff
@@ -162,6 +170,472 @@ async def test_register_agents_and_launch(
...
+    with pytest.raises(ValueError, match='from body'):
+        await body()
+    assert len(manager.running()) == 0
+    # Body raised before submit; handles were never bound.
+    for handle in handles:
+        assert handle._agent_id is None
+
+
+@pytest.mark.asyncio
+async def test_launch_batch_empty_body_does_not_call_register_agents(
```
AK2000: Remove

tests/unit/manager_test.py
AK2000: Never mind, this is caught by a different test below, so I think we should remove this test

tests/unit/manager_test.py
```diff
@@ -162,6 +170,472 @@ async def test_register_agents_and_launch(
...
+    # intent to launch first agent succeeds
+    # intent to launch second agent fails to launch
+    assert failing_launch.await_count == 2  # noqa: PLR2004
+    # Rollback is run once for the orphaned agent left by
+    # the failing launch.
+    assert failing_terminate.await_count == 1
+
+
+@pytest.mark.asyncio
+async def test_launch_batch_propagates_cancellation(
```
AK2000: I don't think this is really testing anything? The cancellation is caught and we aren't checking any cleanup (because no cleanup is done)? I think this can/should be removed?

tests/unit/manager_test.py
```diff
@@ -162,6 +170,472 @@ async def test_register_agents_and_launch(
...
+    # Second launch's handle was bound (agent_id was set) but the
+    # launch call failed; the cache entry was popped during rollback.
+    assert failed._agent_id is not None
+    assert failed._agent_id not in manager._handles
+    # Third launch was never reached: the handle is still unbound.
+    assert orphaned._agent_id is None
+
+
+@pytest.mark.asyncio
+async def test_launch_batch_orphan_termination_failure_does_not_mask(
```
AK2000: From one of the above comments, I think this test should be removed

tests/unit/manager_test.py
```diff
@@ -162,6 +170,472 @@ async def test_register_agents_and_launch(
...
+    h1 = await batch.queue(EmptyAgent)
+    h2 = await batch.queue(EmptyAgent)
+    await batch.submit()
+
+    assert h1.agent_id in manager.running()
+    assert h2.agent_id in manager.running()
+
+
+@pytest.mark.asyncio
+async def test_launch_batch_abort_discards_intents(
```
AK2000: If we remove abort, this can be removed as well

tests/unit/manager_test.py
```diff
@@ -162,6 +170,472 @@ async def test_register_agents_and_launch(
...
+    manager: Manager[LocalExchangeTransport],
+) -> None:
+    async with manager.launch_batch() as batch:
+        handle = await batch.queue(EmptyAgent)
+        await batch.submit()
+    assert manager._handles[handle.agent_id] is handle
+
+
+@pytest.mark.asyncio
+async def test_launch_batch_abort_is_idempotent(
```
AK2000: Remove after abort is removed

tests/unit/manager_test.py
```diff
@@ -162,6 +170,472 @@ async def test_register_agents_and_launch(
...
+    manager: Manager[LocalExchangeTransport],
+) -> None:
+    batch = manager.launch_batch()
+    await batch.queue(EmptyAgent)
+    batch.abort()
+    batch.abort()
+
+
+def test_handle_pickle_unbound_raises() -> None:
```
AK2000: Should probably be moved to `handle_test.py`

tests/unit/manager_test.py
```diff
@@ -162,6 +170,472 @@ async def test_register_agents_and_launch(
...
+    handle: Handle[Any] = Handle()
+    with pytest.raises(pickle.PicklingError, match='unbound'):
+        pickle.dumps(handle)
+
+
+@pytest.mark.asyncio
+async def test_launch_batch_fan_out_sibling_handle(
```
AK2000: Remove

tests/unit/manager_test.py
```diff
@@ -162,6 +170,472 @@ async def test_register_agents_and_launch(
...
+        await batch.queue(_FlexAgent, args=(parent,))
+        await batch.queue(_FlexAgent, kwargs={'peer': parent})
+    running = manager.running()
+    assert len(running) == 3  # noqa: PLR2004
+    assert parent.agent_id in running
+    assert manager._handles[parent.agent_id] is parent
+
+
+@pytest.mark.asyncio
+async def test_launch_batch_diamond_sibling_handles(
```
AK2000: Remove

AK2000 (dismissed):
I think all the logic looks solid now. 

I was a little bit critical of some of the tests just because I want to keep the test volume manageable --- I was operating under the assumption that these were generated tests, which means I was looking to see that the tests were actually necessary to validate some part of the expected behavior, and was aggressive about suggesting some tests be removed. If you think that any of the tests that I suggested to remove feel free to keep them. My goal here is to not let any of the code get to the state where we need Claude to maintain it just because there is too much volume, and I fear if we keep going down this path that is where we end up.

academy/handle.py
```diff
@@ -113,6 +114,25 @@ def __init__(
...
+            raise RuntimeError(
+                'Handle is not bound to a registered agent. '
+                'Submit the enclosing batch before reading '
+                'agent_id.',
+            )
+        return self._agent_id
+
+    @agent_id.setter
+    def agent_id(self, value: AgentId[AgentT_co]) -> None:
+        self._agent_id = value
```
benclifford, line 135: maybe interesting to assert that self._agent_id is None here -- double binding an agent shouldn't even happen in this model and my gut kinda says it would be useful later on when something gets screwed up. doesn't block merging this PR though.

academy/handle.py
```diff
@@ -146,18 +166,20 @@ def __reduce__(
...
             )
-        return (Handle, (self.agent_id,))
+        if self._agent_id is None:
+            raise PicklingError('Cannot pickle an unbound handle.')
+        return (Handle, (self._agent_id,))
 
     def __repr__(self) -> str:
         return (
-            f'{type(self).__name__}(agent_id={self.agent_id!r}, '
+            f'{type(self).__name__}(agent_id={self._agent_id!r}, '
```
benclifford, line 176: from an observability perspective, we now have these handles floating around that you might log out that now have an identity that is richer than just the agent they point to. Someone might want to correlate a handle that points to None to the same handle a bit that that now points to some agent. That suggests adding in some handle identity here in addition to agent identity, for example `id(self)` would capture some of that (although not be fully unique over time).

another thing that shouldn't block merging this PR.

academy/manager.py
benclifford: agree

benclifford: (this implementation is also buggy, I think - it doesn't properly report that I can't abort if I've already submitted and instead silently ignore me?)

academy/manager.py
```diff
@@ -528,6 +730,39 @@ def get_handle(
         self._handles[agent_id] = handle
         return handle
 
+    def launch_batch(self) -> _BatchLauncher:
+        """Open a batch context that joins multiple agent launches.
```
benclifford: ... "that performs multiple simultaneous agent launches"

?

tests/unit/manager_test.py
```diff
@@ -162,6 +170,472 @@ async def test_register_agents_and_launch(
     assert len(manager.running()) == 2  # noqa: PLR2004
 
 
+@pytest.mark.asyncio
+async def test_launch_batch_empty_is_noop(
+    manager: Manager[LocalExchangeTransport],
+) -> None:
+    async with manager.launch_batch():
+        pass
+    assert len(manager.running()) == 0
```
benclifford, line 182: here you would also expect that no exchange transport interaction happens that causes (e.g.) a globus authorization too?

tests/unit/manager_test.py
```diff
@@ -162,6 +170,472 @@ async def test_register_agents_and_launch(
...
+
+@pytest.mark.asyncio
+async def test_launch_batch_launch_after_close_raises(
+    manager: Manager[LocalExchangeTransport],
+) -> None:
+    batch = manager.launch_batch()
+    async with batch:
+        pass
+    with pytest.raises(RuntimeError, match='batch is closed'):
+        await batch.queue(EmptyAgent)
```
benclifford, line 194: should other calls to a batch be raising errors to tell the user they have done something wrong? I'm not sure.

AK2000 (approved):

PR #409, opened 2026-05-12
Adding Custom Serialization Methods

### LINKED ISSUE #406: Configurable Serialization of Requests/Responses
### Describe the Request

By default ActionRequest arguments and ActionResponse results are serialized using pickle. `pickle` is inherently unsafe and allows arbitrary code to be executed. This is a risk when agents can be shared among users. Also, it is not strictly necessary for many arguments and results. I propose we have an option where arguments/results can be serialized as json, and potentially agents or applications can disable `pickle` serialization either on a handle or when receiving requests.

### Sample Code

```python

```

AK2000:
## Summary

This PR enables a user to choose between using JSON or Pickle to serialize arguments and responses.
I chose to make the serialization method part of the messaging protocol rather than having a "try them until one works" approach that was in [proxystore](https://github.com/proxystore/proxystore/blob/main/proxystore/serialize.py) and [globus compute](https://github.com/globus/globus-compute/tree/96e79f843354dc3382675206523a8aa45b9d2df4/compute_sdk/globus_compute_sdk/serialize). My hope is that by making the serialization method more explicit, we can simplify both the process of serialization and the cognitive load of debugging when something fails. 
This also more easily enabled me to do things like specify specific serialization for responses and exceptions. 
Exceptions in particular led me to make a different change: exceptions raised by academy are not serialized and sent via pickle, but communicated via an exception code. This was so that those exceptions could remain independent of the serialization method desired and could be raised at different points in the process (i.e. the exchange could send a mailbox terminated error the same way regardless of how the client wants to deal with serialization).
Right now serialization is attached to "context", meaning an agent or a client can say I want to serialize things a certain way or only want to allow these deserializers. I think a natural extension would be to allow serialization customization at the level of actions as well.

!! Warning
This PR would cause incompatibilities between the hosted exchange and older clients.

Some other features that I want to implement but am leaving to a different PR:
 - Advertising of serialization method on the exchange (i.e. before you send a request you know how the request "should" be serialized).
 - Including pedantic as a serializer to allow action arguments to be validated (i.e. defining arguments and results as pedantic classes the using pedantic methods for serialization)
 - a "raw binary" mode that allows the agent or client to handle serialization internally.

## Related Issues

- Closes #406 

## Changes

## Testing

N/A

## Pull Request Checklist

### DIFF
--- a/academy/exception.py
+++ b/academy/exception.py
@@ -189,8 +189,8 @@ class ExchangeClientNotFoundError(Exception):
 class DeserializationMethodProhibitedError(Exception):
     """Deserialization prohibited.
 
-    Request argument or results are serialized in a method that is prohibtted
-    from being deserialized in the current context.
+    Request argument or results are serialized using a method that is
+    prohibtted from being deserialized in the current context.
     """
 
     def __reduce__(self) -> Any:
@@ -200,7 +200,7 @@ class DeserializationMethodProhibitedError(Exception):
 class ExceptionSerializationError(Exception):
     """Error when serializing an exception.
 
-    An action attempted to return a exception, but that exception could not be
+    An action attempted to return an exception, but that exception could not be
     serialized using the method provided.
     """
 
@@ -221,7 +221,7 @@ class ExceptionSerializationError(Exception):
 class AcademyRemoteError(Exception):
     """Generalized exception for serializing remote exceptions.
 
-    Any exception raised but is then JSON serialized loses its type.
+    Any exception raised that is JSON serialized loses its type.
     To raise the original exception, change `exception_serialization` in the
     action invocation.
     """
--- a/academy/exception.py
+++ b/academy/exception.py
@@ -19,12 +19,11 @@ class ActionCancelledError(Exception):
     configured to cancel running actions.
     """
 
-    def __init__(self, name: str) -> None:
-        super().__init__(f'Action "{name}" was cancelled by the agent.')
-        self.name = name
+    def __init__(self) -> None:
+        super().__init__(f'Action was cancelled by the agent.')
 
     def __reduce__(self) -> Any:
-        return type(self), (self.name,)
+        return type(self), ()
 
 
 class ActionInvalidStateError(Exception):
@@ -186,6 +185,45 @@ class ExchangeClientNotFoundError(Exception):
     def __reduce__(self) -> Any:
         return type(self), (self.aid,)
 
+class DeserializationMethodProhibited(Exception):
+    """Deserialization prohibitted.
+
+    Request argument or results are serialized in a method that is prohibtted
+    from being deserialized in the current context. 
+    """
+
+    def __reduce__(self) -> Any:
+        return type(self), ()
+    
+
+class ExceptionSerializationError(Exception):
+    """Error when serializing an exception.
+
+    An action attepted to return a exception, but that exception could not be
+    serialized using the method provided.
+    """
+
+    def __init__(self, exception_name: str, serializer: str) -> None:
+        super().__init__(
+            f'Exception {exception_name} could not be serialized using '
+            f'method {serializer}. Either change `exception_serialization` in'
+            'the action invocation, or ensure the exception can be '
+            'serialized.'
+        )
+        self.exception_name = exception_name
+        self.serializer = serializer
+
+    def __reduce__(self) -> Any:
+        return type(self), (self.exception_name, self.serializer)
+    
+
+class RemoteException(Exception):
+    """Generalized exception for serializing remote exceptions.
+
+    Any exception raised but is then JSON serialized loses its type. 
+    To raise the original exception, change `exception_serialization` in the
+    action invocation.
+    """
 
 def raise_exceptions(
     exceptions: Iterable[BaseException],
--- a/academy/exception.py
+++ b/academy/exception.py
@@ -20,7 +20,7 @@ class ActionCancelledError(Exception):
     """
 
     def __init__(self) -> None:
-        super().__init__(f'Action was cancelled by the agent.')
+        super().__init__('Action was cancelled by the agent.')
 
     def __reduce__(self) -> Any:
         return type(self), ()
@@ -185,21 +185,22 @@ class ExchangeClientNotFoundError(Exception):
     def __reduce__(self) -> Any:
         return type(self), (self.aid,)
 
-class DeserializationMethodProhibited(Exception):
-    """Deserialization prohibitted.
+
+class DeserializationMethodProhibitedError(Exception):
+    """Deserialization prohibited.
 
     Request argument or results are serialized in a method that is prohibtted
-    from being deserialized in the current context. 
+    from being deserialized in the current context.
     """
 
     def __reduce__(self) -> Any:
         return type(self), ()
-    
+
 
 class ExceptionSerializationError(Exception):
     """Error when serializing an exception.
 
-    An action attepted to return a exception, but that exception could not be
+    An action attempted to return a exception, but that exception could not be
     serialized using the method provided.
     """
 
@@ -208,22 +209,23 @@ class ExceptionSerializationError(Exception):
             f'Exception {exception_name} could not be serialized using '
             f'method {serializer}. Either change `exception_serialization` in'
             'the action invocation, or ensure the exception can be '
-            'serialized.'
+            'serialized.',
         )
         self.exception_name = exception_name
         self.serializer = serializer
 
     def __reduce__(self) -> Any:
         return type(self), (self.exception_name, self.serializer)
-    
 
-class RemoteException(Exception):
+
+class AcademyRemoteError(Exception):
     """Generalized exception for serializing remote exceptions.
 
-    Any exception raised but is then JSON serialized loses its type. 
+    Any exception raised but is then JSON serialized loses its type.
     To raise the original exception, change `exception_serialization` in the
     action invocation.
     """
+
 
 def raise_exceptions(
     exceptions: Iterable[BaseException],
--- a/academy/exchange/client.py
+++ b/academy/exchange/client.py
@@ -30,7 +30,7 @@ from academy.handle import Handle
 from academy.identifier import AgentId
 from academy.identifier import EntityId
 from academy.identifier import UserId
-from academy.message import ErrorResponse
+from academy.message import ACADEMY_ERROR_CODE, AcademyErrorResponse
 from academy.message import Message
 from academy.message import RequestT_co
 from academy.task import spawn_guarded_background_task
@@ -461,8 +461,12 @@ class UserExchangeClient(ExchangeClient[ExchangeTransportT]):
 
     async def _handle_message(self, message: Message[Any]) -> None:
         if message.is_request():
-            error = TypeError(f'{self.client_id} cannot fulfill requests.')
-            response = message.create_response(ErrorResponse(exception=error))
+            response = message.create_response(
+                AcademyErrorResponse(
+                    error_code=ACADEMY_ERROR_CODE.INVALID_CLIENT,
+                    mailbox_id=self.client_id,
+                )
+            )
             await self._transport.send(response)
             logger.warning(
                 'Exchange client for %s received unexpected request message '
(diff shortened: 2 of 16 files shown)


benclifford:
generally looks good. I'll make lots of small comments.

benclifford:
for the to-do list at the end, also consider what it would be like to use a not-pickle serializer by default, with the model that pickle is for "special advanced use only" for users who understand its limitations.

academy/runtime.py
```diff
@@ -262,25 +294,35 @@ async def _execute_action(self, request: Message[ActionRequest]) -> None:
...
             )
         else:
-            response = request.create_response(
-                ActionResponse(result=result),
-            )
             logger.debug(
-                'Completed action %s with invocation id %s',
+                (
+                    'Completed action %s with invocation id %s, result '
+                    'serializer: %s'
```
benclifford, line 322: remember to add interesting fields to the machine readable extras too

academy/exchange/client.py
```diff
@@ -30,7 +30,8 @@
 from academy.identifier import AgentId
 from academy.identifier import EntityId
 from academy.identifier import UserId
-from academy.message import ErrorResponse
+from academy.message import AcademyErrorCode
```
benclifford: its a bit weird repeating the namespace again inside the class name -- that's what namespaces are for?

is there a driver for this?

academy/message.py
```diff
@@ -168,29 +198,113 @@ def get_result(self) -> Any:
...
+    def get_exception(self) -> Exception:
+        """Get the exception.
+
+        Returns:
+            The exception.
+        """
+        match self.error_code:
+            case AcademyErrorCode.MAILBOX_TERMINATED:
+                assert self.mailbox_id is not None, (
+                    'Improper error response created.'
```
benclifford: weird to use the verb "created" here, I think, because the creation was long ago and far away? (this is happening at decode time on a different machine?)

academy/exception.py
```diff
@@ -19,12 +19,11 @@ class ActionCancelledError(Exception):
     configured to cancel running actions.
     """
 
-    def __init__(self, name: str) -> None:
-        super().__init__(f'Action "{name}" was cancelled by the agent.')
-        self.name = name
+    def __init__(self) -> None:
+        super().__init__('Action was cancelled by the agent.')
```
benclifford, line 23: I think this removal of `name` is coming from the flattening of all errors into one structure which doesn't have the necessary fields? (or you don't want to add those fields in).

Which I think is fine for this particular error.

I'm wondering if that will be limiting in future, but its not a serious concern to me right now.

academy/exception.py
```diff
@@ -187,6 +186,47 @@ def __reduce__(self) -> Any:
...
+    """
+
+    def __reduce__(self) -> Any:
+        return type(self), ()
+
+
+class ExceptionSerializationError(Exception):
+    """Error when serializing an exception.
+
+    An action attempted to return a exception, but that exception could not be
```
benclifford: *an* exception, typo

academy/exception.py
```diff
@@ -187,6 +186,47 @@ def __reduce__(self) -> Any:
         return type(self), (self.aid,)
 
 
+class DeserializationMethodProhibitedError(Exception):
+    """Deserialization prohibited.
+
+    Request argument or results are serialized in a method that is prohibtted
```
benclifford: *using* a method.

academy/exception.py
```diff
@@ -187,6 +186,47 @@ def __reduce__(self) -> Any:
...
+        self.serializer = serializer
+
+    def __reduce__(self) -> Any:
+        return type(self), (self.exception_name, self.serializer)
+
+
+class AcademyRemoteError(Exception):
+    """Generalized exception for serializing remote exceptions.
+
+    Any exception raised but is then JSON serialized loses its type.
```
benclifford: fix this grammar

academy/serialize.py
```diff
@@ -1,11 +1,179 @@
...
+        raw = pickle.dumps(obj)
+        return base64.b64encode(raw).decode('utf-8')
+
+    @classmethod
+    def deserialize(cls, data: str) -> Any:
+        """Deserialize data."""
+        return pickle.loads(base64.b64decode(data))
+
+
+class SerializationStrategies(str, Enum):
```
benclifford: this should be a singular noun - e.g. `SerializationStrategy`

academy/serialize.py
```diff
@@ -1,11 +1,179 @@
...
+
+
+default_serializer: ContextVar[SerializationStrategies] = ContextVar(
+    'default_serializer',
+    default=SerializationStrategies.PICKLE,
+)
+"""Default serialization method used to send requests."""
+
+allowed_deserializers: ContextVar[set[SerializationStrategies]] = ContextVar(
+    'deserialization_allow_list',
```
benclifford: make this match the variable name?

docs/guides/persistent.md
```diff
@@ -112,7 +112,7 @@ User=centos
 ExecStart=~/.venv/bin/python -m academy.run --config <config.toml>
 ```
 
-Once you have created you unit file, you can start the service:
+Once you have created a unit file, you can start the service:
```
benclifford, line 115: grumble about unrelated changes

academy/message.py
```diff
@@ -66,8 +94,7 @@ class ActionRequest(BaseModel):
     def _pickle_and_encode_obj(self, obj: Any) -> str:
         if isinstance(obj, str):  # pragma: no cover
             return obj
-        raw = pickle.dumps(obj)
-        return base64.b64encode(raw).decode('utf-8')
+        return serialize(obj, self.serialization)
```
benclifford, line 97: there are three _pickle_and_whatever methods in this file that don't (in general) do pickling any more. I think the name is irrelevant because they're attached to pydantic via decorator, not by name. (or at least, if I mess with the names, tox still passes for me). you could rename them.

academy/message.py
```diff
@@ -168,29 +198,113 @@ def get_result(self) -> Any:
...
-        return base64.b64encode(raw).decode('utf-8')
+        try:
+            return serialize(obj, self.serialization)
+        except Exception:
+            # If we get an exception while serializing an exception,
+            # we do not want to raise an exception and prevent any response
+            # from being returned. Instead we replace the exception with
+            # a exception we know can be serialized, letting the client know
+            # that a exception was hidden.
+            print('Serialization raised eception')
```
benclifford: logs not prints

tests/unit/message_test.py
```diff
@@ -125,15 +169,78 @@ def test_action_response_lazy_deserialize() -> None:
...
+
+
+def test_user_error_response_serialization_error() -> None:
+    response = UserErrorResponse(
+        serialization=SerializationStrategies.PICKLE,
+        exception=UnserializableError(),
+    )
+
+    json = response.model_dump_json()
+    print(json)
```
benclifford: logs not prints. or no prints in tests.

benclifford:
I thought I could grep for `import pickle` and expect to only see it in `academy.serialize` but there is also an import and usage in `academy.message` -- the model deserialize for a message, which is used in a few places in the code. I didn't do a proper audit of when those calls are used, and maybe they're all safe, but the existence of this at all makes me a bit uncomfortable.

academy/handle.py
```diff
@@ -80,20 +82,28 @@ class Handle(Generic[AgentT_co]):
...
         *,
         exchange: ExchangeClient[Any] | None = None,
         ignore_context: bool = False,
+        serializer: SerializationStrategies | None = None,
+        result_serializer: SerializationStrategies | None = None,
+        exception_serializer: SerializationStrategies | None = None,
     ) -> None:
         self._agent_id: AgentId[AgentT_co] | None = agent_id
         self._exchange = exchange
+        self.serialization = serializer
```
benclifford: inconsistent to use serialization vs serializer in these three variables

benclifford: maybe even call it request_serializer to scope it the same as the other two

academy/message.py
```diff
@@ -50,6 +61,23 @@ class ActionRequest(BaseModel):
     """
 
     action: str = Field(description='Name of the requested action.')
+    serialization: SerializationStrategies = Field(
+        description='Serialization strategy used send args',
```
benclifford: missing "to"

academy/message.py
```diff
@@ -136,6 +163,9 @@ class ActionResponse(BaseModel):
     result: SkipValidation[Any] = Field(
         description='Result of the action, if successful.',
     )
+    serialization: SerializationStrategies = Field(
+        description='Serialization strategy used send result.',
```
benclifford: missing "to"

academy/message.py
```diff
@@ -145,16 +175,16 @@ def _pickle_and_encode_result(self, obj: Any) -> list[Any] | None:
...
             isinstance(obj, list)
             and len(obj) == 2  # noqa PLR2004
-            and obj[0] == '__pickled__'
+            and obj[0] == '__serialized__'
         ):  # pragma: no cover
             # Prevent double serialization
             return obj
 
-        raw = pickle.dumps(obj)
+        data = serialize(obj, self.serialization)
```
benclifford, line 184: fix the docstring for this class which refers to this now changed/removed pickling.

academy/runtime.py
```diff
@@ -86,6 +89,11 @@ class RuntimeConfig(BaseModel):
             permanently if the agent shuts down due to an error.
         terminate_on_success: Terminate the agent by closing its mailbox
             permanently if the agent shuts down without an error.
+        default_serializer: Serialization strategy to user when sending
```
benclifford: to use(not-r)

academy/runtime.py
```diff
@@ -238,6 +257,10 @@ async def _execute_action(self, request: Message[ActionRequest]) -> None:
             },
         )
 
+        result_serialization = body.result_serialization or body.serialization
+        exception_serialization = (
+            body.exception_serialization or result_serialization
```
benclifford, line 262: this has the slightly weird behaviour, maybe, that using an empty set (aka *no* permitted deserializers) doesn't actually prohibit all deserializers. not that I expect that to be a practical problem but it doesn't match the docstring 😆

academy/serialize.py
```diff
@@ -1,11 +1,179 @@
...
+        return AcademyRemoteError(dct['exception_str'])
+
+    return dct
+
+
+class JsonSerializer:
+    """Serializes objects into json.
+
+    Works only on basic python types, but allows deserialization without
+    arbitrary code execution, and has increased compatibility between
```
benclifford: increased [with respect to what?]

academy/message.py
```diff
@@ -168,29 +199,113 @@ def get_result(self) -> Any:
...
+        repr=False,
+    )
 
     model_config = DEFAULT_MUTABLE_CONFIG
 
     @field_serializer('exception', when_used='json')
-    def _pickle_and_encode_obj(self, obj: Any) -> str | None:
-        raw = pickle.dumps(obj)
-        return base64.b64encode(raw).decode('utf-8')
+    def _serialize_obj(self, obj: Any) -> str | None:
```
benclifford, line 293: about this pre-existing "when_user" parameter on the decorator -- should this be serializing this way for every way that the UserErrorResponse is serialized? (e.g. if I pickle a UserErrorResponse, I think I would expect the specified serialization to be used, not pickle, and I'm not clear if thats what happens here right now?)

what was happening before this PR? custom pickling when serializing to JSON, and pickle-default-pickling when serializing using pickle?

AK2000:
> ... but there is also an import and usage in academy.message -- the model deserialize for a message, which is used in a few places in the code. I didn't do a proper audit of when those calls are used, and maybe they're all safe, but the existence of this at all makes me a bit uncomfortable.

I thought they were safe, but I changed this to use `model_dump_json` instead. The overhead between dumping as json and pickling should be minor and didn't affect the tests.

academy/message.py
AK2000: The two options for `when_used` are `json` and `python`. When pickling a Pydantic model, none of the `model_dump` functionality is used nor are custom serializers. Every field is just pickled as the actual type of the field and it does not look like any additional validation happens upon unpicking (based off the code here: https://github.com/pydantic/pydantic/blob/ebc4652df990affb79bb3c3db960fb8d2a2b4308/pydantic/main.py#L1156). So both before and after this PR, custom pickling when serializing to JSON and pickle-default-pickling when serializing using pickle.

PR #412, opened 2026-05-18
Adding Protocol Version to Messages

AK2000:
## Summary

Since #406 changed the protocol of the messages sent over the exchange, it has raised the issue that some changes to the network protocol will make it incompatible for agents to talk to each other or for agents to use the hosted exchange. I introduce a protocol_version for messages on the network that we can use to define compatibility beyond just the releases. The idea being a "release" might have some breaking change for code using in the same environment, while a "protocol_version" upgrade may cause agents to be unable to talk to one another. (For instance, agents in v0.3.x and v0.4.0 were able to talk using the same protocol even though other interfaces changed, but this will not be true between agents in v0.4.0 and v0.5.0 because we now require the serialization method as part of the message.)

In the future I expect this versioning to let us think more about what compatibility we support, but at the moment, this is very simple based on major version of the `protocol_version`.

## Related Issues

N/A

## Changes

## Testing

N/A

## Pull Request Checklist

### DIFF
--- a/academy/exception.py
+++ b/academy/exception.py
@@ -232,8 +232,9 @@ class IncompatibleNetworkProtocolError(Exception):
 
     def __init__(self) -> None:
         super().__init__(
-            'Incompatible academy network protocol. This likely means '
-            'you need to upgrade your academy version. ',
+            'Incompatible academy network protocol. The network protocol '
+            'version needs to match between communicating agents and the '
+            'exchange.',
         )
 
     def __reduce__(self) -> Any:
--- a/academy/exception.py
+++ b/academy/exception.py
@@ -230,15 +230,22 @@ class AcademyRemoteError(Exception):
 class IncompatibleNetworkProtocolError(Exception):
     """Received message incompatible with this version."""
 
-    def __init__(self) -> None:
+    def __init__(
+        self,
+        version: str | None = None,
+        cur_version: str | None = None,
+    ) -> None:
         super().__init__(
             'Incompatible academy network protocol. The network protocol '
             'version needs to match between communicating agents and the '
-            'exchange.',
+            f'exchange. Remote version {version}, current version '
+            f'{cur_version}',
         )
+        self.version = version
+        self.cur_version = cur_version
 
     def __reduce__(self) -> Any:
-        return type(self), ()
+        return type(self), (self.version, self.cur_version)
 
 
 def raise_exceptions(
--- a/academy/exception.py
+++ b/academy/exception.py
@@ -227,6 +227,19 @@ class AcademyRemoteError(Exception):
     """
 
 
+class IncompatibleNetworkProtocolError(Exception):
+    """Received message incompatible with this version."""
+
+    def __init__(self) -> None:
+        super().__init__(
+            'Incompatible academy network protocol. This likely means '
+            'you need to upgrade your academy version. ',
+        )
+
+    def __reduce__(self) -> Any:
+        return type(self), ()
+
+
 def raise_exceptions(
     exceptions: Iterable[BaseException],
     *,
--- a/academy/message.py
+++ b/academy/message.py
@@ -39,24 +39,30 @@ from academy.serialize import serialize
 
 DEFAULT_FROZEN_CONFIG = ConfigDict(
     arbitrary_types_allowed=True,
-    extra='forbid',
+    extra='ignore',
     frozen=True,
     use_enum_values=True,
     validate_default=True,
 )
 DEFAULT_MUTABLE_CONFIG = ConfigDict(
     arbitrary_types_allowed=True,
-    extra='forbid',
+    extra='ignore',
     frozen=False,
     use_enum_values=True,
     validate_default=True,
 )
 
-PROTOCOL_VERSION: Version = Version('0.0')
+PROTOCOL_VERSION: Version = Version('1')
 
 
 def check_version(version: str | None) -> bool:
-    """Checks is a `protocol_version` is compatible.
+    """Checks if a `protocol_version` is compatible.
+
+    Messages within a major version are intended to be mutually
+    compatible, while minor version changes might introduce
+    optional fields or remove existing fields without prohibiting
+    their existence. Local tags may be added for branches/forks
+    and must match exactly.
 
     Args:
         version: The version string to check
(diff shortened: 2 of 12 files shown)


academy/exception.py
```diff
@@ -227,6 +227,19 @@ class AcademyRemoteError(Exception):
     """
 
 
+class IncompatibleNetworkProtocolError(Exception):
+    """Received message incompatible with this version."""
+
+    def __init__(self) -> None:
+        super().__init__(
```
yadudoc, line 238: I recommend adding the version info on the exchange to the message. It's useful to know what version to upgrade to for example.

academy/message.py
```diff
@@ -49,6 +52,22 @@
     validate_default=True,
 )
 
+PROTOCOL_VERSION: Version = Version('0.0')
+
+
+def check_version(version: str | None) -> bool:
+    """Checks is a `protocol_version` is compatible.
```
yadudoc:
```suggestion
    """Checks if `protocol_version` is compatible.
```

yadudoc (dismissed):
This looks good overall. I've got a couple of minor questions.

benclifford:
I think none of the discussion here or text in the PR should be using "upgrade": if I'm talking to a centrally hosted service using version 88, and I install an academy client that talks version 89, the solution is not for me to upgrade anything. Nor is it for me to wait a 6-18 month release cycle for the centrally hosted service to upgrade.

academy/message.py
```diff
@@ -49,6 +52,22 @@
     validate_default=True,
 )
 
+PROTOCOL_VERSION: Version = Version('0.0')
```
benclifford: start at version 1. none of this semver "haha I said version 0 so i can away with breaking things, the version number is meaningless".

only use one field, "1", until such time as you have defined the semantics of any other fields. Which might not be a sequential minor version but might look more like SMTP or X509 extension names, for example. the commit message and the code specifically *doesn't* scribe any meaning to that second .0

figure out whats going to happen when that version number contains something that isn't a 2 part version and/or a floating point number and document that now - it's more interesting to say "version 1 is exactly the string "1", version "2" might be "2+featureA+featureB" or "2.4.23.2" but all we're describing is how to check for version 1"

academy/message.py
```diff
@@ -362,6 +384,13 @@ class Header(BaseModel):
         ),
     )
     kind: Literal['request', 'response']
+    protocol_version: str | None = Field(
+        str(PROTOCOL_VERSION),
+        description=(
+            'Version of the academy protocol used. Until version 1.0, the '
```
benclifford: enough of this. if we're cosplaying versions, we should have nice costumes.

academy/message.py
```diff
@@ -439,6 +468,11 @@ def label(self) -> uuid.UUID | None:
         """Message label."""
         return self.header.label
 
+    @property
+    def protocol_version(self) -> str | None:
+        """Message label."""
```
benclifford: this is a cut and paste from above i think - rewrite

examples/04-execution/run-04.py
```diff
@@ -54,7 +54,10 @@ async def main() -> int:
     )
 
     async with await Manager.from_exchange_factory(
-        factory=HttpExchangeFactory(),
+        factory=HttpExchangeFactory(
+            'https://exchange.academy-agents.org/main',
```
benclifford: is this change intentional?

academy/message.py
benclifford: (actually no matter, this is a comment on versioning of academy versions, not the network version, so I'm not concerned)

benclifford:
i'm wondering a bit, but I haven't tried anything out, about what a protocol n+1 can look like but still be able for this version of academy to understand a message enough to give a decent protocol error.

An n+1 needs to be a `Message` in the sense of both academy protocol n+1 and this current version (and so implicitly, also in the sense of intermediate versions), and it needs to have a protocol_version field.

Is it also then going to be required to have src/dest/tag fields (even if protocol n+1 doesn't need them) in order to deserialize enough to get to the protocol checking field?

tests/unit/exchange/client_test.py
```diff
@@ -297,6 +298,71 @@ async def test_client_reply_error_on_request(
...
+        async with await factory.create_agent_client(
+            registration,
+            _request_handler,
+        ) as agent_client:
+            message = Message.create(
+                src=agent_client.client_id,
+                dest=client.client_id,
+                body=PingRequest(),
+            )
+            message.header = message.header.model_copy(
```
benclifford: i'd be more comfortable if this message-from-the-future was more realistic looking: for example, make a message that has a few imagined changes (that don't need to be real...), and check it is still handled properly, rather than changing only the message version.

What are the expected changes? new action types? new fields? removed fields?

academy/message.py
yadudoc: Protocol will most likely require updates, and not all of them will be backward compatible. It is a good idea to encode that into the protocol version number. This info also helps the client route to the right api endpoint, at least more than having an integer version number.

examples/04-execution/run-04.py
yadudoc: Yes, this points to a separate exchange instance that runs the latest code from main.

AK2000: Ah, no, this was not supposed to go into this commit, I'll revert it

AK2000:
> i'm wondering a bit, but I haven't tried anything out, about what a protocol n+1 can look like but still be able for this version of academy to understand a message enough to give a decent protocol error.
> 
> An n+1 needs to be a `Message` in the sense of both academy protocol n+1 and this current version (and so implicitly, also in the sense of intermediate versions), and it needs to have a protocol_version field.
> 
> Is it also then going to be required to have src/dest/tag fields (even if protocol n+1 doesn't need them) in order to deserialize enough to get to the protocol checking field?

I believe you're correct, but if we change the message/header to `extra='ignore'`, then we can delete fields that have a default value. In general, I think I support doing this because it will allow more non-breaking (i.e. if we add fields later, a message would still be parsable in this version). The update in `client_test.py` shows an example of a future message I could imagine with a completely new body and extra information in the header, that still create a decent protocol error.

If we find it's necessary later, we could also implement a pre-parsing check on messages to read the protocol version from the raw python/json and return an appropriate error before we use pedantic to parse to `Message`

academy/message.py
benclifford: What sort of protocol updates will be not backward compatible, but enough to keep the major version the same with the existing major version check? It would be good to add a test case or two around that sort of expected change.

AK2000: I added a test case around this. The one I came up with is adding a TTL into an ActionRequest. When it is received by the "old" (read: current) version of the protocol, the TTL is ignored, and when an "old" request is read into the new version, the TTL is assumed to be None.

Issue #324, opened 2026-02-11
Fine Grained Permissions


AK2000:
### Describe the Request

Currently the HTTP Exchange only implements access based on tokens. If an agent can access the exchange, they can talk to all agents belonging to that user. We want more fine-grained permissions where agent can only read from the mailbox they have been assigned, and the ability to pass messages can be restricted to a smaller et of mailboxes.

### Sample Code

```python

```

PR #416, opened 2026-05-22
Auth friction and sub-agents convo

GusEllerm:
## Summary

This PR is a place to discuss and decide on how we want to 1. reduce auth friction from users using hosted exchanges (Globus), and 2. enable agents to delegate auth to sub-agents. This initial PR is part of a proposal I have, but does not lock us into any particular solution. 

Launching agents on the Globus Exchange today requires users to *authenticate twice*, and agents *cant spawn sub-agents*. My proposed solution is described below to get us started:

Agents already have Globus Auth identities, and per-agent scope registration, but those identities are not used at runtime. The agent uses a user delegated token to access the exchange, so every action runs through the users auth chain. We redesign this so those existing identities are used by agents to authenticate as themselves via `client_credentials`, so the token presented to the exchange is native to the agent, not the user. After a one-time provisioning, agents boot non-interactively, persist across sessions, and can spawn other agents on demand. 

You could imagine this -- initial provisioning of some agent client:
```python
async with await Manager.from_exchange_factory(factory=factory) as m:
    creds = await m.register_persistent_agent(MyAgent)
    # creds.agent_id is durable; credentials persisted at
    # $ACADEMY_HOME/agents/<uid>.json with 0o600 perms.
```
and reuse of that client avoiding re-auth:
```python
# Booting — any future session, any process, zero prompts:
async with await Manager.from_exchange_factory(factory=factory) as m:
    creds = store.load(agent_id)  
    handle = await m.launch(
        MyAgent,
        registration=PersistentGlobusAgentRegistration(creds),
    )
```
and then an agent with sub-agent launch capabilities:
```python
class Coordinator(Agent):
    @action
    async def deploy_worker(self) -> AgentId[Worker]:
        return await self.spawn_child(Worker, name='worker-1')
        # Parent's own client_credentials grant mints child credentials.
        # No user prompt; no user session needed.
```

This functionality requires:

1. Agents authenticating themselves via `client_credentials`, instead of as the user's delegate. This removes one auth request. 
2. Agent mailboxes are shared with a _fleet Globus Group_ when provisioned for auth
3. Each agent's credentials (`client_id`, `secret`) are persisted on disk so they can boot non-interactively. This is a pre-req for sub-agent spawning + long-horizon agents. 
4. Parent agents provision children using their own `manage_projects` --> hierarchical relationships between agents are represented in the exchange. 

All four are validated end-to-end against the live Globus dev exchange by some spike scripts, so in theory this should work.  

This specific PR relates to point 3. @AK2000 already pushed a PR (#397) which persists credentials on disk for long-lived agents via the `python -m academy.run` path. This PR adds similar infrastructure: a storage convention for agent credentials, plus lookup and boot primitives. The flow `agent_id` -> `credentials_map` -> `re-launched agent` works today, but there is **no** user facing flow that exercises it yet (future PR). 

Identity mapping between agents and agent-credentials is done via UUID. `PersistentAgentCredentials.agent_id` carries the agents UUID. The store writes agents/{agent_id.uid}.json. The file path is the identity mapping. 

---

### Globus Groups as auth

Right now an agent is a thin layer on top of the user's identity. The user logs in, delegates a token, and the agent uses that token to access the exchange on the user's behalf. 

The idea is to make each agent its own [Globus Confidential Client](https://globus-sdk-python.readthedocs.io/en/stable/_modules/globus_sdk/services/auth/client/confidential_client.html), authenticating against the exchange via `client_credentials` OAuth grant. Once an agent is running under its own auth, the user's session is not required. The agent could run indefinitely, its credentials can be written to disk, and reloaded across restarts. 

We can also make an agent **spawn-capable** at its registration, which would (in this proposal) grant it the `manage_projects` capability, enabling it to mint creds for child agents. In this model a user is required to provision a parent-agent and the fleet of agents under that parent can grow without additional auth prompts. 

This all assumes that we move the auth model from per-agent scopes to fleet groups ([Globus Groups](https://www.globus.org/globus-groups-service)). Currently the exchange does this check: "does this token bear scope `agent_scope`", which requires a consent prompt per agent. With groups this becomes "is this agent a member of fleet group X?". So per-agent scopes are not required. 

The logic for this [already exists](https://github.com/academy-agents/academy/blob/056a898014a9a55d31765a036c8db5f2770fadbd/academy/exchange/cloud/authenticate.py#L196-L211), and authz check runs in the backend (look for group_memberships in the [backend](https://github.com/academy-agents/academy/blob/main/academy/exchange/cloud/backend.py#L283-L291)).  Its used for **explicit** mailbox sharing (e.g. a user chooses to share a mailbox with a group), so this proposal is to make this the default rather than opt-in, and to develop the wiring needed to manage fleet memberships.  

### Secondary wins

Relationship disambiguation on the exchange. As each agent has its own client credentials and authenticates as itself, we can resolve an agents `client_id` at the exchange. So, the exchange can now know which agent sent a message, and we can create attestation / provenance chains. You could enforce things like only a parent shutting down its child, only a fleet member invoking a fleet action, or rejecting a message whose claimed src doesn't match  the authenticated sender.

## Related Issues

- #45 — This PR's credential store is the foundation for the agent_id identity persistence that issue calls for 
- #324 — the plan set out here uses group-based authz for mailbox access, which is a step toward the narrower per-agent permission model.

## Changes

- [X] Enhancement (non-breaking change or feature addition)
- [X] Refactor (internal code or design clean up)
- [X] Test (changes or additions to testing)

## Testing

New unit tests included in PR.

## Pull Request Checklist

- [X] Tests have been added to show the fix is effective or that the new feature works.
- [X] Docs have been updated and reviewed if relevant.

### DIFF
--- a/academy/exchange/cloud/credentials.py
+++ b/academy/exchange/cloud/credentials.py
@@ -40,13 +40,7 @@ class PersistentAgentCredentials(BaseModel, Generic[AgentT]):
     """Client ID of the agent's Globus Auth resource server."""
 
     client_secret: SecretStr
-    """Secret used to authenticate the agent's Globus Auth client.
-
-    Stored as ``SecretStr`` so ``repr`` and accidental ``model_dump_json``
-    calls do not leak the secret. The credential file store opts into
-    unmasking via ``model_dump_json(context={'unmask': True})``; all other
-    callers see ``'**********'``.
-    """
+    """Secret used to authenticate the agent's Globus Auth client."""
 
     scope_string: str
     """Per-agent scope requested via ``client_credentials``."""
@@ -64,8 +58,7 @@ class PersistentAgentCredentials(BaseModel, Generic[AgentT]):
     """Whether this agent may spawn child agents.
 
     Default ``False``. ``True`` grants project-admin authority and is
-    not safe for production until sub-project isolation lands in v1.1
-    (see threat model T1).
+    not safe for production until isolation is implemented. 
     """
 
     model_config = ConfigDict(frozen=True)
@@ -153,8 +146,7 @@ class AgentCredentialFileStore:
         """List identifiers for all agents with credentials on disk.
 
         Returns ``AgentId`` instances with ``name=None``. Call
-        ``load(agent_id)`` if the display name is needed; this method
-        deliberately avoids reading every file just to enumerate IDs.
+        ``load(agent_id)`` if the display name is needed.
         """
         return [
             AgentId(uid=uuid.UUID(p.stem))
--- a/academy/exchange/cloud/credentials.py
+++ b/academy/exchange/cloud/credentials.py
@@ -1,0 +1,170 @@
+"""Persistent agent credentials and on-disk store.
+
+Used by the Globus persistent-agent boot path to authenticate as the
+agent via ``client_credentials`` without further user interaction.
+"""
+
+from __future__ import annotations
+
+import logging
+import os
+import pathlib
+import uuid
+from datetime import datetime
+from typing import Any
+from typing import Generic
+
+from pydantic import BaseModel
+from pydantic import ConfigDict
+from pydantic import field_serializer
+from pydantic import SecretStr
+from pydantic import SerializationInfo
+
+from academy.exchange.cloud.login import get_academy_home
+from academy.identifier import AgentId
+from academy.identifier import AgentT
+
+logger = logging.getLogger(__name__)
+
+_AGENTS_SUBDIR = 'agents'
+_AGENTS_DIR_MODE = 0o700
+
+
+class PersistentAgentCredentials(BaseModel, Generic[AgentT]):
+    """Durable credentials for a Globus persistent agent."""
+
+    agent_id: AgentId[AgentT]
+    """Unique identifier for the agent."""
+
+    client_id: uuid.UUID
+    """Client ID of the agent's Globus Auth resource server."""
+
+    client_secret: SecretStr
+    """Secret used to authenticate the agent's Globus Auth client.
+
+    Stored as ``SecretStr`` so ``repr`` and accidental ``model_dump_json``
+    calls do not leak the secret. The credential file store opts into
+    unmasking via ``model_dump_json(context={'unmask': True})``; all other
+    callers see ``'**********'``.
+    """
+
+    scope_string: str
+    """Per-agent scope requested via ``client_credentials``."""
+
+    agent_type: tuple[str, ...]
+    """MRO of the agent type, matching ``Agent._agent_mro()``."""
+
+    fleet_group_ids: tuple[uuid.UUID, ...]
+    """Fleet group memberships at provision time (length 1 in v1)."""
+
+    created_at: datetime
+    """Provisioning timestamp."""
+
+    spawn_capable: bool = False
+    """Whether this agent may spawn child agents.
+
+    Default ``False``. ``True`` grants project-admin authority and is
+    not safe for production until sub-project isolation lands in v1.1
+    (see threat model T1).
+    """
+
+    model_config = ConfigDict(frozen=True)
+
+    @field_serializer('client_secret', when_used='json')
+    def _serialize_client_secret(
+        self,
+        secret: SecretStr,
+        info: SerializationInfo,
+    ) -> str:
+        if info.context is not None and info.context.get('unmask'):
+            return secret.get_secret_value()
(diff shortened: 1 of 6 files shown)


academy/exchange/cloud/credentials.py
```diff
@@ -0,0 +1,162 @@
...
+    """Fleet group memberships at provision time (length 1 in v1)."""
+
+    created_at: datetime
+    """Provisioning timestamp."""
+
+    spawn_capable: bool = False
+    """Whether this agent may spawn child agents.
+
+    Default ``False``. ``True`` grants project-admin authority and is
+    not safe for production until isolation is implemented.
```
benclifford, line 61: is there a sensible path for that implementation?

academy/exchange/cloud/login.py
```diff
@@ -50,6 +50,21 @@ def prompt_for_code(self) -> str:  # pragma: no cover
         return auth_code.strip()
 
 
+def get_academy_home() -> pathlib.Path:
```
benclifford, line 53: put this somewhere more general and pull it out of this PR? at least the json pool logger also wants this sort of thing:

$ git grep share.acad
academy/logging/configs/jsonpool.py:    `~/local/share/academy/logs/`

academy/exchange/cloud/credentials.py
```diff
@@ -0,0 +1,162 @@
...
+
+    def __init__(
+        self,
+        directory: pathlib.Path | None = None,
+    ) -> None:
+        if directory is None:
+            directory = get_academy_home() / _AGENTS_SUBDIR
+        self._dir = pathlib.Path(directory)
+        self._dir.mkdir(parents=True, exist_ok=True)
+        self._dir.chmod(_AGENTS_DIR_MODE)
```
benclifford, line 92: theres probably some security relevant race condition here between the mkdir and chmod.

eg some of the hits on https://www.google.com/search?q=mkdir+chmod+toctou

academy/exchange/cloud/credentials.py
```diff
@@ -0,0 +1,162 @@
...
+    """Secret used to authenticate the agent's Globus Auth client."""
+
+    scope_string: str
+    """Per-agent scope requested via ``client_credentials``."""
+
+    agent_type: tuple[str, ...]
+    """MRO of the agent type, matching ``Agent._agent_mro()``."""
+
+    fleet_group_ids: tuple[uuid.UUID, ...]
+    """Fleet group memberships at provision time (length 1 in v1)."""
```
benclifford, line 52: what is this for? it only seems to be referenced in tests?

academy/exchange/cloud/credentials.py
benclifford: and who gets to do what now with this? and how is that restricted when "isolation is implemented"?

academy/exchange/cloud/credentials.py
GusEllerm: Right -- thats an artefact from part of the future implementation. Not best practice to include here without its functionality, but it helps explain the direction. The idea is to have group memberships handle auth checks on agent mailboxs (replacing per-agent scopes). I updated the PR comment with more info re this direction. If the proposal looks good after some convo I can pull this to make the PRs more self contained.

academy/exchange/cloud/credentials.py
GusEllerm: Yes -- for each spawn-capable agent you create a sub-project. First we enable agents to be project admins for the user's main Globus project (dangerous, but a stepping stone). If creds leak an attacker basically has full access. I assume that users trust themselves, and we expose an explicit warning if `spawn_capable=True` regarding that risk. 

Then we can work on isolation. Each spawn-capable agent owns a child Globus Project minted when they are provisioned. `Manager.register_persistent_agent(...., spawn_capable=True)` calls `auth_client.create_project(...)`, which creates the parent client within a sub-project as admin. Children are minted in the sub-project. `PersistentAgentCredentials` gets a `project_id` so parents know which project to create children in. 

So that isolation restricts auth actions, but agents can still communicate. Sub-projects bind what an agent can create or destroy in Globus. Coms is seperate -- each agent the user launches holds the same exchange scope, so they all reach the same exchange, and fleet group membership decides whose mailbox they can read.

benclifford:
note from a private conversation:

> Another question on your proposal is about what this unlocks in terms of other auth. For example do we now get a trust chain where an agent can believe the remote agent id, enough for auth, if restricting who can call actions

PR #436, opened 2026-07-21
Custom executor to launch agents in same event loop.

### LINKED ISSUE #321: Run Multiple Agents in Same Event Loop
### Describe the Request

I want to be able to launch multiple agents in a single event loop that is not the same event loop as the manager. This Is useful for instance for launching many /O bound agents --- but the manager does not live in the same place the agents need to run. You can hack around this using the `agent.launch_along_side` functionality, but this is only available within the agent.

### Sample Code

```python

```

jz1909:
## Summary
Create new custom executor to run multiple agents in one event loop

## Related Issues
 - Closes #321 

## Changes

- [!] Test (changes or additions to testing)

## Testing
N/A and in progress

## Pull Request Checklist

- [In Progress] Code changes pass `pre-commit` (e.g., ruff, mypy, etc.).
- [In Progress] New and existing unit tests pass locally with the changes.
- [IN Progress] Docs have been updated and reviewed if relevant.

### DIFF
--- a/academy/executor.py
+++ b/academy/executor.py
@@ -1,6 +1,10 @@
 from __future__ import annotations
 
 import asyncio
+import contextlib
+
+import academy.exchange as ae
+
 from collections.abc import Callable
 from typing import Any
 from concurrent.futures import Executor
@@ -10,8 +14,9 @@ from academy.agent import action
 from academy.agent import Agent
 from academy.handle import Handle
 from academy.manager import _run_agent_on_worker, _RunSpec
-import academy.exchange as ae
 from academy.runtime import RuntimeConfig
+from academy.exception import AgentTerminatedError
+
 
 
 
@@ -31,8 +36,6 @@ class _EventLoopHost(Agent):
         """Run a submitted Callable in this agent's own event loop.
         Args:
             fn: Callable function to run, where an awaitable result is awaited before returning.
-            args: Position args for fn.
-            kwargs: Keyword args for fn.
 
         Returns:
         Results of fn and awaited potentially.=
@@ -78,8 +81,6 @@ class EventLoopExecutor(Executor):
 
         Args: 
             fn: Callable to run on the host.
-            Args: Positional args for fn.
-            Kwargs: Keyword args for fn.
 
         Returns: Future resolving when fn finishes on host.
         """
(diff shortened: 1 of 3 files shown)


benclifford:
Academy has a 'Agent.agent_launch_alongside` which allows an agent to launch another agent in the current event loop. This PR doesn't use any of that logic, but is doing something quite similar.

I'd expect any similar functionality to either use that call, merge that previous implementation into the new code here, or justify why that can't be done.

jz1909:
> Academy has a 'Agent.agent_launch_alongside` which allows an agent to launch another agent in the current event loop. This PR doesn't use any of that logic, but is doing something quite similar.
> 
> I'd expect any similar functionality to either use that call, merge that previous implementation into the new code here, or justify why that can't be done.

Looking at the latest commit for this PR, I think agent_launch_alongside and event_loop_executor serve two different enough purposes and event_loop_executor can't directly utilize agent_launch_alongside

Agent_launch_alongside is for the use case where the user wants to make child agents helpers of the host agent, where there is an explicit relationship between them. This makes sense especially since the user have to manually create a parent agent for this to work

however, in executor, we dispatch a new thread and hide the host_agent, which changes the use case from hierarchical relationships to instead wanting multiple agents running parallel on one worker.

in the event loop executor case, the host agent doesn't know about the child agent, as they're registered independently and the host agent only acts to run the child agent as a callable, but in the agent_launch_alongside case, the parent agent explicitly gets a handle, contains the agent_id, and has a directly relationship to it

We can't just use agent_launch_alongside in event_loop_executor in a trivial way because it lacks the ability to go outside the caller's event loop and make a separate loop in a separate thread.

academy/executor.py
```diff
@@ -0,0 +1,201 @@
...
+            fn: Callable to run on the host.
+            *args: Positional arguments for submitted function.
+            **kwargs: Keyword arguments for submitted function.
+
+        Returns: Future resolving when fn finishes on host.
+        """
+        if self._shutdown:
+            raise RuntimeError('Cannot submit after host shutdown')
+
+        if fn is _run_agent_on_worker:
```
AK2000, line 104: I think this needs to be an assert or we need to raise an error, I don't like failing in some nondescript way if we get a function that we don't expect. This executor is only designed to work for this specific function and will not work otherwise, right?

AK2000 (dismissed):

PR #446, opened 2026-08-13
Re-Authentication on Forbidden Error

AK2000:
## Summary

We are starting to run into scenarios of long running workflows where token expiry is a significant issue. The previous instructions were to use the `GlobusExchangeFactory`, which minted agents as Globus resource servers and would automatically use a refresh token to deal with token expiration. The issue is that new agents in that model always required an authorization approval, making that an infeasible option for workflows that create agents on-the-fly.

To deal with token expiration, when a forbidden error is returned,  I retry after fetching a (new) token from the globus authenticator.  This will work anywhere with access to a refresh token or access to client credentials. This should be the case when there is a shared filesystem between the manager and the compute nodes, or when a user has previously authenticated from that machine. If a new token cannot be fetched, the retry will throw an error. This is not as robust as the `GlobusExchangeFactory`, but removes significant barriers --- hopefully we can find a better in between.

Alternatives Considered:
 - Turning the `HttpExchangeFactory` into a `globus_sdk.BaseClient` like the `GlobusExchangeClient`. This is much bigger shift and the `globus_sdk` is all synchronous code, leading to lots of threading ugliness.
 - Shipping the refresh token over the wire at agent launch. Would ensure the refresh token is available to re-authenticate, but replicating and reusing refresh tokens is not good practice since refresh tokens never expire.
 - Creating a single resource server that all agents (of a user) use as for credentials. Then a single delegated token could serve all agents and we wouldn't have to re-auth. 

## Related Issues

N/A

## Changes

## Testing

N/A

## Pull Request Checklist

### DIFF
--- a/academy/exchange/cloud/client.py
+++ b/academy/exchange/cloud/client.py
@@ -23,6 +23,7 @@ else:  # pragma: <3.11 cover
     from typing_extensions import Self
 
 import aiohttp
+from aiohttp import ClientResponse
 from aiohttp import hdrs
 from pydantic import BaseModel
 from pydantic import Field
@@ -60,7 +61,7 @@ DEFAULT_EXCHANGE_URL = 'https://exchange.academy-agents.org/v1'
 
 class _HttpConnectionInfo(NamedTuple):
     url: str
-    additional_headers: dict[str, str] | None = None
+    additional_headers: dict[str, str]
     ssl_verify: bool | None = None
     request_timeout_s: float = 60
     client_timeout: aiohttp.ClientTimeout | None = None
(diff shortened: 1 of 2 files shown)


academy/exchange/cloud/client.py
```diff
@@ -344,33 +417,27 @@ class HttpExchangeConsole:
     Args:
         session: Http session.
         connection_info: Exchange connection info.
+        auth_method: Authenticataion method used by exchange.
```
benclifford:
```suggestion
        auth_method: Authentication method used by exchange.
```

PR #447, opened 2026-08-14
Option to Reject Actions when Agent is Inactive

AK2000:
## Summary

There are many times where you do not want to put a message in the mailbox of an agent that's not online (for instance the agent died unexpectedly and you expect it to be online, the agent may be functioning properly but you don't want to wait for it to be online, etc.). Building off heartbeats (#366) and the new inactive status (#404), this PR adds an option to a handle to  raise an error on the client side calling the action if the agent is "INACTIVE". For this PR, we poll the exchange separately from the message listening loop.

## Related Issues

 - Not an open issue, but this came up in the StuctBioReasoner work 

## Changes

## Testing

N/A

## Pull Request Checklist

### DIFF
--- a/academy/exchange/mailbox_status.py
+++ b/academy/exchange/mailbox_status.py
@@ -1,0 +1,16 @@
+from __future__ import annotations
+
+import enum
+
+
+class MailboxStatus(enum.Enum):
+    """Exchange mailbox status."""
+
+    MISSING = 'MISSING'
+    """Mailbox does not exist."""
+    ACTIVE = 'ACTIVE'
+    """Mailbox exists and is accepting messages."""
+    INACTIVE = 'INACTIVE'
+    """Mailbox accepting messages but has missed heartbeats."""
+    TERMINATED = 'TERMINATED'
+    """Mailbox was terminated and no longer accepts messages."""
--- a/academy/handle.py
+++ b/academy/handle.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import asyncio
+import contextlib
 import functools
 import logging
 import time
@@ -14,8 +15,10 @@ from typing import TYPE_CHECKING
 from typing import TypeVar
 from weakref import WeakSet
 
+from academy.exception import AgentInactiveError
 from academy.exception import AgentTerminatedError
 from academy.exception import ExchangeClientNotFoundError
+from academy.exchange.transport import MailboxStatus
 from academy.identifier import AgentId
 from academy.message import ActionRequest
 from academy.message import ActionResponse
@@ -79,6 +82,8 @@ class Handle(Generic[AgentT_co]):
     Args:
         agent_id: ID of the remote agent, or ``None`` to defer
             binding (used by ``batch.queue()``).
+        reject_on_inactive: Return an error when target agent is inactive.
+            Needs to be a positional argument for serialization.
         exchange: A default exchange client to be used if an exchange client
             is not configured in the current context.
         ignore_context: Ignore the current context and force use of `exchange`
@@ -90,6 +95,7 @@ class Handle(Generic[AgentT_co]):
             the same strategy as the request serializer.
         exception_serializer: Strategy used to serialize results. If false-y,
             use the same strategy as the result serializer.
+        polling_interval: Interval to poll exchange to see if agent is active.
 
     Raises:
         ValueError: If `ignore_context=True` but `exchange` is not provided.
@@ -98,12 +104,14 @@ class Handle(Generic[AgentT_co]):
     def __init__(  # noqa: PLR0913
         self,
         agent_id: AgentId[AgentT_co] | None = None,
+        reject_on_inactive: bool = False,
         *,
         exchange: ExchangeClient[Any] | None = None,
         ignore_context: bool = False,
         request_serializer: SerializationStrategy | None = None,
         result_serializer: SerializationStrategy | None = None,
         exception_serializer: SerializationStrategy | None = None,
+        polling_interval: float = 60,
     ) -> None:
         self._agent_id: AgentId[AgentT_co] | None = agent_id
         self._exchange = exchange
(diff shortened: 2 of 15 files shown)


academy/handle.py
```diff
@@ -190,11 +227,12 @@ def __reduce__(
 
         return (
             Handle,
-            (self._agent_id,),
+            (self._agent_id, self.reject_on_inactive),
```
benclifford: this breaks pickleability sending handles to v0.5.0 - maybe thats something to care about, maybe not.

academy/handle.py
```diff
@@ -126,10 +130,21 @@ def __init__(  # noqa: PLR0913
...
+        self._pending_actions: set[uuid.UUID] = set()
+        self._pending_actions_lock = asyncio.Lock()
 
         if self._exchange is not None:
             self._register_with_exchange(self._exchange)
 
+        # _agent_status is the private variable to keep track of the mailbox
+        # status in the polling loop --- it is not always current (based on
+        # when the last poll was), and is only updated if reject_on_inactive
+        # is true. For live mailbox stuts use `agent_status`
```
benclifford:
```suggestion
        # is true. For live mailbox stats use `agent_status`
```

academy/handle.py
```diff
@@ -430,6 +551,12 @@ async def ping(self, *, timeout: float | None = None) -> float:
         if isinstance(body, ErrorResponse):
             raise body.get_exception()
 
+        if self.reject_on_inactive:
+            # A successful ping indicates active regaurdless of
```
benclifford:
```suggestion
            # A successful ping indicates active regardless of
```

academy/handle.py
```diff
@@ -224,15 +262,90 @@ async def remote_method_call(*args: Any, **kwargs: Any) -> R:
...
-        """
+        """Return live exchange-level metrics for the remote agent."""
         return await self.exchange.agent_stats(self.agent_id)
 
+    async def agent_status(self) -> MailboxStatus:
+        """Return live status of the agent mailbox."""
+        status = await self.exchange.status(self.agent_id)
+        if self._reject_on_inactive:
+            # To avoid a situation where agent_status shows returns
+            # active, and actions are rejected with a stale status
```
benclifford, line 275: I think this TOCTOU race condition can happen anyway: after a user requests agent status, and before they invoke an action, the status polling loop can execute and change the current status so that actions will be rejected.

AK2000: The worry is not that we reject an action after `agent_status` returns `ACTIVE`, but that we reject the action based on stale information. If the polling loop runs after `agent_status` and returns `INACTIVE` that means the agent was marked `INACTIVE` between the `agent_status` call and the `action` call. I think rejecting the action is the right behavior.

academy/handle.py
```diff
@@ -224,15 +262,90 @@ async def remote_method_call(*args: Any, **kwargs: Any) -> R:
...
+
+            await asyncio.sleep(self.polling_interval)
+
+    async def _notify_pending_actions(self) -> None:
+        """Notify pending actions when agent becomes inactive."""
+        loop = asyncio.get_event_loop()
+
+        # Typically self._agent_status == INACTIVE should act as a lock, but
+        # while we are rejecting actions, the agent could become active
+        # and new actions could be added to the pending_actions set.
```
benclifford, line 307: related to this, for an agent which repeatedly toggles between ACTIVE and INACTIVE, this code will send a cancel for an action repeatedly, every time that INACTIVE happens, up until the point that a response message comes back, right? I guess that's fine though.

tests/unit/handle_test.py
```diff
@@ -687,3 +689,165 @@ async def test_handle_exception_serializer(
...
+                MailboxStatus.INACTIVE,
+                MailboxStatus.INACTIVE,
+            ]
+            registration = await client.register_agent(SleepAgent)
+            handle = Handle(
+                registration.agent_id,
+                polling_interval=TEST_SLEEP_INTERVAL,
+            )
+            handle.reject_on_inactive = True
+            await handle._handle_ready.wait()
```
benclifford: is this event really a barrier/flag that self._agent_status has had a real value set in it? (rather than some more abstract notion of the entire handle being ready)

academy/exchange/mailbox_status.py
```diff
@@ -0,0 +1,16 @@
+from __future__ import annotations
+
+import enum
+
+
+class MailboxStatus(enum.Enum):
```
benclifford, line 6: this move is another API breakage to document in migrations.md

academy/handle.py
```diff
@@ -79,6 +78,8 @@ class Handle(Generic[AgentT_co]):
     Args:
         agent_id: ID of the remote agent, or ``None`` to defer
             binding (used by ``batch.queue()``).
+        reject_on_inactive: Return an error when target agent is inactive.
+            Needs to be a positional argument for serialization.
```
benclifford: why is that? because you want it to be set via the property accessor methods in `__init__` rather than set by `__setstate__` directly into the dictionary?

academy/handle.py
```diff
@@ -150,6 +165,28 @@ def agent_id(self, value: AgentId[AgentT_co]) -> None:
...
+            # Start loop if needed
+            self._handle_ready = asyncio.Event()
+            self._reject_on_inactive = value  # Set after event to avoid race
+            self._status_task: asyncio.Task[None] = asyncio.create_task(
+                self._status_loop(),
+                name=f'{self.handle_id}-status-loop',
+            )
+        elif old and not value:
+            # Stop loop if not needed
+            self._reject_on_inactive = value  # Set before canel to avoid race
```
benclifford:
```suggestion
            self._reject_on_inactive = value  # Set before cancel to avoid race
```

academy/handle.py
```diff
@@ -224,15 +262,90 @@ async def remote_method_call(*args: Any, **kwargs: Any) -> R:
...
+
+    async def _notify_pending_actions(self) -> None:
+        """Notify pending actions when agent becomes inactive."""
+        loop = asyncio.get_event_loop()
+
+        # Typically self._agent_status == INACTIVE should act as a lock, but
+        # while we are rejecting actions, the agent could become active
+        # and new actions could be added to the pending_actions set.
+        async with self._pending_actions_lock:
+            for response_tag in self._pending_actions:
```
benclifford: these tags are the initial request tags from the action submission, right? (the tags being in a response anywhere is not so relevant?)

academy/handle.py
AK2000: This was before I had put the starting and stopping the loop into the setter as well. So now we don't have this issue. I can fix the comment and move this into a  key-word only arg

academy/handle.py
AK2000: Good catch, it's not an issue sending the message, but it is an issue to set an exception on a future multiple times.

